PR #30, opened 2017-05-26
[Issue 6] merged pravega writer and exactly once writer code

### LINKED ISSUE #6: Merge FlinkExactlyOncePravegaWriter and FlinkPravegaWriter
**Problem description**
The FlinkExactlyOncePravegaWriter currently always uses transactions to write to pravega. Hence if this is used as a sink in a flink job with checkpointing disabled the events will never be committed to pravega.
We could conditionally use transactions in the writer depending on whether checkpointing is enabled or not.

**Problem location**
FlinkExactlyOncePravegaWriter 

Merge the 2 writer implementations.

vijikarthi:
**Change log description**
Combines `FlinkPravegaWriter` and `FlinkExactlyOncePravegaWriter` code to offer a common entry point for writing events to Pravega streams.   Closes #6.

**Purpose of the change**
`FlinkExactlyOncePravegaWriter` uses Pravega transactions and it works only when checkpoint is enabled. When job is configured to use `FlinkExactlyOncePravegaWriter` and checkpoint is not enabled, the writes will not be committed.

**What the code does**
- Combines `FlinkPravegaWriter` and `FlinkExactlyOncePravegaWriter`
- Introduces `EXACTLY_ONCE` as a new option for `PravegaWriterMode`
- Introduces an internal writer abstraction

**How to verify it**
- Integration Tests: `FlinkPravegaWriterTest` and `FlinkExactlyOncePravegaWriterTest`


EronWright:
This PR is jarring to read thru.  There's so much conditional logic here, almost begging for some refactoring.   One cannot tell which variables are used in which mode(s).

I suggest defining an `InternalWriter' inner class, with a subclass for the transactional path and another for the non-transactional path.  Some of the variables could be moved accordingly.   This may help to reduce the conditional logic and organize the state.

vijikarthi:
I started with that approach in mind but backed off since the object initialization needs to happen on both open() as well as restoreState() calls to query the checkpoint configuration status. Anyway, I will go ahead and make the change that you have suggested as it improves the readability.

I would prefer to have the user explicitly set ATLEAST_ONCE or EXTACTLY_ONCE options on the writer instead of the code trying to infer from the pipeline for two reasons. 
1) This will give more flexibility to the users who wanted to use ATLEAST_ONCE writer capabilities but still have checkpoint enabled for the job (highly unlikely configuration but probable scenario) 
2) Users need to be aware that transaction and checkpoint are inter-linked.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -55,106 +73,464 @@
...
+    private final long txnTimeoutMillis;
+    private final long txnMaxTimeMillis;
+    private final long txnGracePeriodMillis;
 
-    // Following runtime parameters are populated when the flink sub-tasks are executed at the task managers.
+    // Writer interface to assist exactly-once and atleast-once functionality
+    private transient InternalWriter<T> writerImpl = null;
 
-    // The pravega writer client.
+    /** The pravega writerImpl client. Used only to create transactions. */
```
EronWright: Looks like you (or your IDE) was overzealous in renaming `writer` to `writerImpl`.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -55,106 +73,464 @@
...
+        this(controllerURI, scope, streamName, serializationSchema, router,
+                DEFAULT_TXN_TIMEOUT_MILLIS, DEFAULT_TXN_TIMEOUT_MILLIS, DEFAULT_TX_SCALE_GRACE_MILLIS);
+    }
 
-    // Thread pool for handling callbacks from write events.
-    private transient ExecutorService executorService = null;
 
     /**
-     * The flink pravega writer instance which can be added as a sink to a flink job.
+     * The flink pravega writerImpl instance which can be added as a sink to a flink job.
```
EronWright: `writer` vs `writerImpl`

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -55,106 +73,464 @@
...
-                this.streamName,
-                this.eventSerializer,
+                streamName,
+                serializer,
                 EventWriterConfig.builder().build());
-        this.executorService = Executors.newFixedThreadPool(5);
-        log.info("Initialized pravega writer for stream: {}/{} with controller URI: {}", this.scopeName,
-                 this.streamName, this.controllerURI);
+
+        log.info("Initialized pravega writerImpl for stream: {}/{} with controller URI: {}", scopeName,
```
EronWright: `writer` vs `writerImpl`

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -174,37 +550,47 @@ public void invoke(T event) throws Exception {
...
-    public void snapshotState(FunctionSnapshotContext context) throws Exception {
-        if (this.writerMode == PravegaWriterMode.ATLEAST_ONCE) {
+        @Override
+        public void close() throws Exception {
+            flushAndVerify();
+            pravegaWriter.close();
+        }
+
+        @Override
+        public List<UUID> snapshotState(long checkpointId, long checkpointTime) throws Exception {
```
EronWright: Given that this writer isn't used when snapshots are enabled, is this method dead code?

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -55,106 +73,464 @@
...
+
+
+    /**
+     * Restores the state, which here means the IDs of transaction for which we have
+     * to ensure that they are really committed.
+     */
+    @Override
+    public void restoreState(List<UUID> list) throws Exception {
+        // restore will be called before open. Create a local internal writer only to assist the restore functionality
+        InternalWriter<T> writerImpl;
```
EronWright: Smells bad to use a local writerImpl variable when a field of the same name exists, and to use an uninitialized writer instance.   It also seems odd that the writerImpl isn't mutated by the call to `restoreState`.

I can think of a few alternatives:
1. Move the writerImpl-related code from `open` to a shared `initializeWriter` method that would also be called from `restoreState`.   
2. use a static method on `TransactionWriter` to restore state and assert that checkpointing is enabled.   This would make most sense if `restoreState` doesn't mutate the writerImpl's state.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -55,106 +73,464 @@
...
+
+        /** The currently running transaction to which we write */
+        private transient Transaction<T> currentTxn;
+
+        /** The transactions that are complete from Flink's view (their checkpoint was triggered),
+         * but not fully committed, because their corresponding checkpoint is not yet confirmed */
+        private transient ArrayDeque<TransactionAndCheckpoint<T>> txnsPendingCommit;
+
+        @Override
+        public void initialize() throws Exception {
```
EronWright: Maybe you should move the logic in `initialize` to the writerImpl constructor.   Then you could mark `txnsPendingCommit` as final.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -55,106 +73,464 @@
...
+
+    }
+
+    /*
+     * Exactly-Once Pravega Writer implementation
+     */
+    private class TransactionWriter implements InternalWriter<T> {
+
+        /** The currently running transaction to which we write */
+        private transient Transaction<T> currentTxn;
```
EronWright: No need to mark `currentTxn` or `txnsPendingCommit` as transient, since they're no longer part of the serializable parent class.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -174,37 +550,47 @@ public void invoke(T event) throws Exception {
...
         }
-    }
 
-    @Override
-    public void initializeState(FunctionInitializationContext context) throws Exception {
-        // Nothing to restore.
-    }
+        @Override
+        public void notifyCheckpointComplete(long checkpointId) throws Exception {
+            //do nothing
```
EronWright: Maybe we should throw `UnsupportedOperationException` in these checkpoint methods.   Or are you intending for `StandardWriter` to be used in some checkpoint scenarios?    How practical would it be to have checkpointing enabled but turn off transaction mode?   Maybe WRITE_MODE=AT_LEAST_ONCE should be the determinant, as opposed to isCheckpointEnabled.

src/test/java/io/pravega/connectors/flink/FlinkPravegaWriterTest.java
```diff
@@ -150,7 +149,7 @@ private void runTest(final int jobParallelism, final boolean withFailure, final
                 setParallelism(jobParallelism);
         execEnv.setRestartStrategy(RestartStrategies.fixedDelayRestart(1, 0));
 
-        DataStreamSource<Integer> dataStream = execEnv.enableCheckpointing(1000, CheckpointingMode.EXACTLY_ONCE)
```
EronWright: Evidently the `FlinkPravegaWriterTest` was testing the non-transactional writer with snapshotting enabled, since it had some code (in `snapshotState`) for that mode.

StephanEwen:
Is there ever a case when you would want to use "at-least-once" mode when you have checkpoints enabled?

I would argue that there is, for example if you are interested in low-latency messaging (milliseconds), where you cannot wait for the checkpoint interval to commit the transactions.

I would suggest that this writer takes an enum that defines it as `AT_LEAST_ONCE´ or `EXACTLY_ONCE`.

fpj:
pravega/pravega#1213 will enable rebalances without checkpoints. Rebalances currently are tied to checkpoints and once we enabled them without checkpoints, we should be able to have what you are asking for. Does it work for you, @StephanEwen ?

StephanEwen:
@fpj Are you talking about Flink checkpoints or Pravega checkpoints here?

fpj:
That´s Pravega checkpoints.

EronWright:
You'll hit a minor conflict with the writer, as we updated it to use the latest pravega client.

vijikarthi:
I have pulled the latest changes and updated the code to merge Standard and Exactly Once writer to single writer. Transaction API will be used only when explicitly asked for using `PravegaWriterMode.EXACTLY_ONCE` option and checkpointing is enabled. Please take a look.

chrisdail:
LGTM

EronWright:
I would like a chance to go thru this before merging.

EronWright:
@vijikarthi I rebased this onto the dev-0.2 branch (since the master branch was stabilized for Pravega 0.1).

I made some minor adjustments to better encapsulate the internal writer's state.   It now handles opening and closing the Pravega `EventStreamWriter`.   I also noticed and fixed that the `ClientFactory` and `ExecutorService` weren't being closed.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -117,48 +164,411 @@ public void setPravegaWriterMode(PravegaWriterMode writerMode) {
         this.writerMode = writerMode;
```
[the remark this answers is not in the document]
EronWright: We don't expect it to change, and maybe the constructor should have an overload for it.   There's a stylistic aspect; the connector fits into Flink's DSL, which acts rather like a builder.

I was planning to have a follow-up issue for improving on the DSL.  I'd like the setter to return `this` to allow for a builder style, and introduce setters for the transaction timeouts.   In other words, I am leaning away from overloaded constructors.  Ultimately we should be consistent with the Flink DSL.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -117,48 +164,411 @@ public void setPravegaWriterMode(PravegaWriterMode writerMode) {
...
+
+    }
+
+    /*
+     * Implements an {@link AbstractInternalWriter} using transactional writes to Pravega.
+     */
+    private class TransactionalWriter extends AbstractInternalWriter {
+
+        /** The currently running transaction to which we write */
+        private Transaction<T> currentTxn;
```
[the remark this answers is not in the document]
EronWright: I too was concerned about that, but reviewed it and think not.   Flink guarantees single-threaded access to the function.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -117,48 +164,411 @@ public void setPravegaWriterMode(PravegaWriterMode writerMode) {
...
+        @Override
+        public void open() throws Exception {
+            // start the transaction that will hold the elements till the first checkpoint
+            this.currentTxn = pravegaWriter.beginTxn();
+            log.debug("{} - started first transaction '{}'", name(), this.currentTxn.getTxnId());
+        }
+
+        @Override
+        public void write(T event) throws Exception {
+            this.currentTxn.writeEvent(eventRouter.getRoutingKey(event), event);
```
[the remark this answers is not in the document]
EronWright: The higher-level lifecycle of a sink function makes it impossible that it would be null, but I agree that an assertion would be handy and will add one.   The transaction cannot be created any earlier.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -117,48 +164,411 @@ public void setPravegaWriterMode(PravegaWriterMode writerMode) {
...
+                log.info("{} - checkpoint {} complete, committing completed checkpoint transaction {}",
+                        name(), checkpointId, txn.transaction().getTxnId());
+
+                // the big assumption is that this now actually works and that the transaction has not timed out, yet
+
+                // TODO: currently, if this fails, there is actually data loss
+                // see https://github.com/pravega/flink-connectors/issues/5
+                txn.transaction().commit();
+
+                log.debug("{} - committed checkpoint transaction {}", name(), txn.transaction().getTxnId());
```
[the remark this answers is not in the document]
EronWright: I am wondering if a race condition could happen.   Could a subsequent transaction overtake the one that is still committing?

PR #33, opened 2017-06-19
Usability improvements for authoring Flink+Pravega apps

chrisdail:
**Change log description**

Improvements to building flink+pravega apps. Changes include:

- Moved Deserialization/Serialization classes from samples into the connector. Users can always provide their own but these default implementations allow getting started easier. Both of our examples copy/pasted these.
- Added convenience class for creating JavaSerializer with the above serialization schema classes. This simplifies code using these.
- Start to establishing conventions around parameters. Expect a parameter named 'controller' to contain the URI to the controller. This convention will make it easier to build apps that run on Nautilus. (I'm open to changing the name if people would prefer a different value. I was also thinking 'pravegaController').
- Added convention for describing Pravega scope and stream name as [scope]/[stream].
- Added utility class for getting stream from a parameter.

**Purpose of the change**

Making it easier to author Flink Pravega apps and easier to integrate with other tools.

For example. This is existing code from the samples:

```
FlinkPravegaReader<Event> flinkPravegaReader = new FlinkPravegaReader<>(URI.create(controllerUri), scope,
        Sets.newHashSet(stream),
        startTime,
        new PravegaDeserializationSchema<>(Event.class, new JavaSerializer<Event>()),
        readerName);
```

Using the new helper it would look like this:
```
FlinkPravegaReader<Event> flinkPravegaReader = pravega.newReader(getStreamId(), startTime, Event.class);
```

**How to verify it**

This makes no change to the connector itself. A separate PR will update the pravega-samples to make use of these changes.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
+++ b/src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *   
+ */
+package io.pravega.connectors.flink.serialization;
+
+import io.pravega.client.stream.Serializer;
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.ByteBuffer;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.streaming.util.serialization.AbstractDeserializationSchema;
+
+/**
+ * A deserialization schema adapter for a Pravega serializer.
+ */
+public class PravegaDeserializationSchema<T extends Serializable> extends AbstractDeserializationSchema<T> {
+    private final Class<T> typeClass;
+    private final Serializer<T> serializer;
+
+    public PravegaDeserializationSchema(Class<T> typeClass, Serializer<T> serializer) {
+        this.typeClass = typeClass;
+        this.serializer = serializer;
+    }
+
+    @Override
+    public T deserialize(byte[] message) throws IOException {
+        ByteBuffer msg = ByteBuffer.wrap(message);
+        return serializer.deserialize(msg);
+    }
+
+    @Override
+    public TypeInformation<T> getProducedType() {
+        return TypeInformation.of(typeClass);
+    }
+}
--- a/src/main/java/io/pravega/connectors/flink/util/FlinkPravegaParams.java
+++ b/src/main/java/io/pravega/connectors/flink/util/FlinkPravegaParams.java
@@ -25,7 +25,7 @@ import org.apache.flink.streaming.util.serialization.DeserializationSchema;
 import org.apache.flink.streaming.util.serialization.SerializationSchema;
 
 /**
- * Convience class for working with pravega parameters coming in as parameters.
+ * Convenience class for extracting pravega parameters from flink job parameters.
  *
  * As a convention, the pravega controller uri will be passed as a parameter named 'controller'.
  *
--- a/src/main/java/io/pravega/connectors/flink/util/FlinkPravegaParams.java
+++ b/src/main/java/io/pravega/connectors/flink/util/FlinkPravegaParams.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ */
+package io.pravega.connectors.flink.util;
+
+import com.google.common.collect.Sets;
+import io.pravega.client.admin.StreamManager;
+import io.pravega.client.stream.ScalingPolicy;
+import io.pravega.client.stream.StreamConfiguration;
+import io.pravega.connectors.flink.FlinkPravegaReader;
+import io.pravega.connectors.flink.FlinkPravegaWriter;
+import io.pravega.connectors.flink.PravegaEventRouter;
+import io.pravega.connectors.flink.serialization.PravegaSerialization;
+import java.io.Serializable;
+import java.net.URI;
+import org.apache.flink.api.java.utils.ParameterTool;
+import org.apache.flink.streaming.util.serialization.DeserializationSchema;
+import org.apache.flink.streaming.util.serialization.SerializationSchema;
+
+/**
+ * Convience class for working with pravega parameters coming in as parameters.
+ *
+ * As a convention, the pravega controller uri will be passed as a parameter named 'controller'.
+ *
+ * Operations are provided to create streams, readers and writers based on stream names as
+ * parameters.
+ *
+ * @see StreamInfo
+ */
+public class FlinkPravegaParams {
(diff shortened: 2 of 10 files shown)


src/main/java/io/pravega/connectors/flink/util/FlinkPravegaParams.java
```diff
@@ -0,0 +1,131 @@
...
+import io.pravega.connectors.flink.PravegaEventRouter;
+import io.pravega.connectors.flink.serialization.PravegaSerialization;
+import java.io.Serializable;
+import java.net.URI;
+import org.apache.flink.api.java.utils.ParameterTool;
+import org.apache.flink.streaming.util.serialization.DeserializationSchema;
+import org.apache.flink.streaming.util.serialization.SerializationSchema;
+
+/**
+ * Convience class for working with pravega parameters coming in as parameters.
```
vroy: Typo in `Convience`. 

I think `pravega parameters coming in as parameters` could be improved.

chrisdail: Fixed

vroy:
LGTM

src/main/java/io/pravega/connectors/flink/util/FlinkPravegaParams.java
```diff
@@ -0,0 +1,131 @@
...
+ * Convenience class for extracting pravega parameters from flink job parameters.
+ *
+ * As a convention, the pravega controller uri will be passed as a parameter named 'controller'.
+ *
+ * Operations are provided to create streams, readers and writers based on stream names as
+ * parameters.
+ *
+ * @see StreamInfo
+ */
+public class FlinkPravegaParams {
```
chrisdail: I'm not 100% sold on the class name here. I welcome alternatives

EronWright:
We should get signoff from Pravega team for the defaults being encoded here.   e.g.:
- the `scope/stream` format
- the default scaling policy of `fixed(1)`

src/main/java/io/pravega/connectors/flink/util/FlinkPravegaParams.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Gets the controller URI from the 'controller' job parameter. If this parameter
+     * is not specified this defaults to DEFAULT_CONTROLLER_URI.
+     */
+    public URI getControllerUri() {
+        return URI.create(params.get(CONTROLLER_PARAM_NAME, DEFAULT_CONTROLLER_URI));
+    }
+
+    /**
+     * Gets a stream information from a stream specification.
```
EronWright: The javadoc here is a bit confusing, it gets stream information from a job parameter.  Maybe `streamParam` should be `streamParamName` too.

src/main/java/io/pravega/connectors/flink/util/StreamInfo.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package io.pravega.connectors.flink.util;
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Captures the fully qualified name of a stream. The convention to represent this as a
+ * single string is using [scope]/[stream].
+ */
+public class StreamInfo {
```
EronWright: `StreamId`  might be more specific than `StreamInfo`

src/test/java/io/pravega/connectors/flink/util/FlinkPravegaParamsTest.java
```diff
@@ -0,0 +1,46 @@
...
+            new String[] { "--controller", "tcp://controller.pravega.l4lb.thisdcos.directory:9091"}
+        ));
+        assertEquals(URI.create("tcp://controller.pravega.l4lb.thisdcos.directory:9091"), helper.getControllerUri());
+    }
+
+    @Test
+    public void testStreamParam() {
+        String input = "testScope/exampleStream";
+        FlinkPravegaParams helper = new FlinkPravegaParams(ParameterTool.fromArgs(
+                new String[] { "--input-stream", input}
```
EronWright: Would you consider using`--input` as a parameter name?  I think it nicer.

chrisdail:
@EronWright I hear your comment about Java serialization vs what Flink supports. I think there is some work required on the Pravega side to better integrate here. Whatever is generating the events needs to support the same serialization. The only thing built into the client today is the JavaSerializer.

I'm open to other options here. I was really more trying to make things easier to construct.

src/main/java/io/pravega/connectors/flink/util/FlinkPravegaParams.java
chrisdail: Will change this.

src/main/java/io/pravega/connectors/flink/util/StreamInfo.java
chrisdail: I like `StreamId`, will change to this.

src/test/java/io/pravega/connectors/flink/util/FlinkPravegaParamsTest.java
chrisdail: This is really just a test and not advocating a specific parameter. People may look at it for an example though so I will change this to `input`. I like the idea of using `input` as the convention if there is a single stream.

src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.ByteBuffer;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.streaming.util.serialization.AbstractDeserializationSchema;
+
+/**
+ * A deserialization schema adapter for a Pravega serializer.
+ */
+public class PravegaDeserializationSchema<T extends Serializable> extends AbstractDeserializationSchema<T> {
```
fpj: There is nothing really specific to Pravega in this class, should we give it a more generic name? I'm also a bit confused by the use of the term `schema` here. Why are we calling it schema?

src/main/java/io/pravega/connectors/flink/util/StreamId.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package io.pravega.connectors.flink.util;
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Captures the fully qualified name of a stream. The convention to represent this as a
+ * single string is using [scope]/[stream].
+ */
+public class StreamId {
```
fpj: This utility class is clearly useful, but I'm wondering whether this is useful specifically to this project or useful more broadly. We have the `stream` interface (and corresponding impl) in Pravega, but it doesn't provide a `fromSpec` call. Is this something we should consider having in Pravega directly rather than have applications create their own?

src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
chrisdail: This class name comes from the sample projects. It is the same as was used there. It is not specific to pravega but it does use Pravega's Serializer interface.

The term "schema" is a flink thing. They call them Deserialization/Serialization Schemas. I don't love the term either but it does implement the "AbstractDeserializationSchema" flink class so the name does make sense. I think it is best to be consistent with Flinks terminology here.

src/main/java/io/pravega/connectors/flink/util/StreamId.java
chrisdail: This is your call if you want to move it into the core of pravega. I think it is more generally useful. If it does move though, it will raise the question if some of the core APIs should be changed to use this ID class instead of scope and name as individual parameters.

chrisdail: I think what is best here is to commit this as is. If Pravega adopts this as part of the core code, we can then change this to use that class instead of having its own.

PR #40, opened 2017-07-02
Usability improvements for PravegaDeserializationSchema

StephanEwen:
**Change log description**

Adds various small improvements to the `PravegaDeserializationSchema`:

  - Avoid redundant double-wrapping of serializers in FlinkPraveaReader
  - Adds a check that the Pavega serializer is serializable (otherwise this gives an exception on execution).
  - Fully support Flink's type information spectrum (also generic types that need to be captured)
  - Add more convenience constructors

**How to verify it**

Checkstyle and unit tests pass.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -9,6 +9,7 @@
  */
 package io.pravega.connectors.flink;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 
 import io.pravega.client.ClientFactory;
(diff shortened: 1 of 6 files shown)


src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -228,7 +229,9 @@ public void run(SourceContext<T> ctx) throws Exception {
                 getRuntimeContext().getTaskNameWithSubtasks(), readerId, this.controllerURI);
 
         // create the adapter between Pravega's serializers and Flink's serializers
-        final Serializer<T> deserializer = new FlinkDeserializer<>(this.deserializationSchema);
+        final Serializer<T> deserializer = this.deserializationSchema instanceof PravegaDeserializationSchema ?
```
chrisdail: Maybe instead of checking for an instance of PravegaDeserializationSchema we can introduce a new interface (that this class would implement) that implements the method:

```public Serializer<T> getSerializer()```

This way someone who is using their own serialization schema can make use of this same code if they implement that interface.

StephanEwen: That sounds like a good idea...

StephanEwen:
I addressed the comments and fixed another issue related to the serializers.

I also removed the `Serializable` scope for generic types, because the code handles also non-serializable types.

Issue #45, opened 2017-08-25
Implement 'fixed' event router


EronWright:
**Problem description**
To use the Pravega sink, an instance of `PravegaEventRouter` must be provided.   Develop a predefined implementation, similar to `FlinkFixedPartitioner`, that maps each Flink partition to a routing key.

Consider whether it would be possible to support parallelism changes to the sink.

_This approach is one of many that could be imagined to make sense of the ordering of events produced to a stream by a Flink program.  A program may ensure that all events for a given key arrive at the same sink instance by leveraging a keyed stream, e.g. `stream.keyBy(...).addSink(..)`.  Some event-time reordering logic would also be needed to establish a per-key ordering._

See [`org.apache.flink.streaming.connectors.kafka.partitioner.FlinkFixedPartitioner`](https://ci.apache.org/projects/flink/flink-docs-release-1.3/api/java/org/apache/flink/streaming/connectors/kafka/partitioner/FlinkFixedPartitioner.html) for background.

**Suggestions**
- Evolve the `PravegaEventRouter` interface with an `open(int parallelInstanceId, int parallelInstances)` method.

EronWright:
I filed this because I don't fully understand the behavior of keyed streams in combination with sinks.  For example, in the following program, would all events for a given key be routed to the same parallel sink instance?  

```scala
val keyedStream = env.addSource(...).keyBy(_.sensorId).setParallelism(3);
keyedStream.addSink(new FlinkPravegaWriter(..., router = _.sensorId)).setParallelism(3);
```

StephanEwen:
When you do a `keyBy()` all the same keys are in one parallel Flink task. If they originate from different sources, then they are in the order in which the network transports them.

To make sure they all end up in the same partition in Pravega, I assume that we would always need a router, because across recoveries, the readers are probably not sticky to a segment.

StephanEwen:
Is your suggestion to plugin a default router if the stream is keyed from Flink's perspective?

Unless you want events strictly ordered by event time, I think routing by key in the FlinkPravegaWriter alone is actually sufficient.

EronWright:
@StephanEwen please have a look at PR #46 to advance this discussion.

EronWright:
Closing this thanks to PR #46 which satisfies my goal of providing a rational ordering of elements per routing key.

PR #46, opened 2017-08-30
Support for event time ordering of elements produced by the sink

EronWright:
**Change log description**
Support for event time ordering of elements produced by the sink.
Closes issue #45.

**Purpose of the change**
This PR provides a facility for reordering a program's output events into event time order, in combination with the natural ordering guarantee of Pravega routing keys.   It works by combining a keyed stream, a custom operator that reorders elements, and a forwarding strategy to the normal sink.

**What the code does**
The code provides three new classes:
- `EventTimeOrderingOperator` - a generic operator that reorders events into event time (using the watermark to flush event buffers).
- `PravegaEventRouterKeySelector` - a helper class to use a pravega event router as the key selector for a Flink keyed stream.
- `FlinkPravegaUtils` - a convenience method to assemble the topology.

**How to verify it**
Unit tests are provided - `EventTimeOrderingOperatorTest`.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/util/FlinkPravegaUtils.java
+++ b/src/main/java/io/pravega/connectors/flink/util/FlinkPravegaUtils.java
@@ -1,0 +1,34 @@
+package io.pravega.connectors.flink.util;
+
+import io.pravega.connectors.flink.EventTimeOrderingOperator;
+import io.pravega.connectors.flink.FlinkPravegaWriter;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.datastream.DataStreamSink;
+
+public class FlinkPravegaUtils {
+
+    /**
+     * Writes a stream of elements to a Pravega stream with event time ordering.
+     * <p>
+     * This method returns a sink that automatically reorders events by their event timestamp.  The ordering
+     * is preserved by Pravega using the associated routing key.
+     *
+     * @param stream      the stream to read.
+     * @param writer      the Pravega writer to use.
+     * @param parallelism the degree of parallelism for the writer.
+     * @param <T>         The type of the event.
+     * @return a sink.
+     */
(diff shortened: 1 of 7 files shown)


EronWright:
Note that this code is largely cribbed from Flink CEP ([ref](https://github.com/apache/flink/blob/master/flink-libraries/flink-cep/src/main/java/org/apache/flink/cep/operator/AbstractKeyedCEPPatternOperator.java)).   The CEP library reorders elements internally to correctly perform temporal pattern matching.  The reordering logic seems highly generic and especially useful here, to be able to reason about the ordering of events in a stream as produced by a Flink program.

I'm not thrilled by the name of the utility method `writeToPravegaInEventTimeOrder`, open to suggestions.

src/main/java/io/pravega/connectors/flink/util/FlinkPravegaUtils.java
```diff
@@ -0,0 +1,43 @@
...
+     * This method returns a sink that automatically reorders events by their event timestamp.  The ordering
+     * is preserved by Pravega using the associated routing key.
+     *
+     * @param stream      the stream to read.
+     * @param writer      the Pravega writer to use.
+     * @param parallelism the degree of parallelism for the writer.
+     * @param <T>         The type of the event.
+     * @return a sink.
+     */
+    public static <T> DataStreamSink<T> writeToPravegaInEventTimeOrder(DataStream<T> stream, FlinkPravegaWriter<T> writer, int parallelism) {
```
EronWright: @StephanEwen does this seem legit to you, to force a forwarding strategy.   Striving to ensure that all elements for a given key are written by a specific sink instance.

EronWright: Note that the Kafka producer has a similarly complex `writeTo` method that composes an internal operator with the sink operator.   For different reasons.  Just pointing out the similarity.
https://github.com/apache/flink/blob/release-1.3.2/flink-connectors/flink-connector-kafka-0.10/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer010.java#L118

PR #49, opened 2017-09-08
Updated to work with latest pravega

jonny-miller:
Moved mavenLocal to the top of the repo list and removed references to AckFuture to compile against latest pravega

**Change log description**
Moved mavenLocal to the top of the repos
Changed AckFuture references to CompletableFuture and changed `addListener` to `thenRunAsync`

**Purpose of the change**
Fix the build to work with the latest pravega

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
@@ -10,7 +10,6 @@
 package io.pravega.connectors.flink;
 
 import io.pravega.client.ClientFactory;
-import io.pravega.client.stream.AckFuture;
 import io.pravega.client.stream.EventStreamWriter;
 import io.pravega.client.stream.EventWriterConfig;
 import io.pravega.client.stream.Serializer;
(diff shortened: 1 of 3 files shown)


src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -160,12 +160,12 @@ public void invoke(T event) throws Exception {
         }
 
         this.pendingWritesCount.incrementAndGet();
-        final AckFuture ackFuture = this.pravegaWriter.writeEvent(this.eventRouter.getRoutingKey(event), event);
+        final CompletableFuture future = this.pravegaWriter.writeEvent(this.eventRouter.getRoutingKey(event), event);
         if (writerMode == PravegaWriterMode.ATLEAST_ONCE) {
-            ackFuture.addListener(
+            future.thenRunAsync(
```
jonny-miller: I am not 100% sure this is the right replacement but it seemed like it might be.

EronWright: I don't think it is because `thenRunAsync` runs only if the future completed normally, not exceptionally.   Meanwhile the catch block seems intended to handle an exception from `writeEvent`. All told, `whenCompleteAsync` makes the most sense here.

Pseudo-code:
```
future.whenCompleteAsync(
                    (result, e) -> {
                        if(e == null) {
                            synchronized (this) {
                                pendingWritesCount.decrementAndGet();
                                this.notify();
                            }
                        }
                        else {
                            log.warn("Detected a write failure: {}", e);

                            // We will record only the first error detected, since this will mostly likely help with
                            // finding the root cause. Storing all errors will not be feasible.
                            writeError.compareAndSet(null, e);
                        }
                    },
                    executorService
            );
```

PR #50, opened 2017-09-08
Setting a port for restserver.

arvindkandhare:
**Change log description**
Sets a port for the rest server.
**Purpose of the change**
This will fix errors when `InProcPravegaCluster` is instantiated during the tests.
**What the code does**
Sets a value for the port.
**How to verify it**
Unit tests do not fail at  `InProcPravegaCluster` initialization.

### DIFF
--- a/src/test/java/io/pravega/connectors/flink/utils/SetupUtils.java
+++ b/src/test/java/io/pravega/connectors/flink/utils/SetupUtils.java
@@ -68,6 +68,7 @@ public final class SetupUtils {
                 .isInMemStorage(true)
                 .isInProcController(true)
                 .controllerCount(1)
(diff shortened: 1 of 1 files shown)


EronWright:
Sorry I don't understand the whole scenario, but could we use a generated port rather than a hardcoded one?   Like we do for the other ports: `int hostPort = TestUtils.getAvailableListenPort();`

EronWright:
Please consider amending the log line to report the endpoint info, if you think a developer might make use of it.
```
         log.info("Initialized Pravega Cluster.");
         log.info("   Controller: tcp://{}:{}/", ...);
         log.info("   REST server: http://{}:{}/", ...);
```

Issue #53, opened 2017-09-26
Exception thrown while trying to connect to controller. 


shiveshr:
**Problem description**

```    java.lang.ArrayIndexOutOfBoundsException: 1
        at io.pravega.client.stream.impl.ControllerResolverFactory.lambda$newNameResolver$103(ControllerResolverFactory.java:68)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
        at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1374)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
        at io.pravega.client.stream.impl.ControllerResolverFactory.newNameResolver(ControllerResolverFactory.java:69)
        at io.grpc.internal.ManagedChannelImpl.getNameResolver(ManagedChannelImpl.java:440)
        at io.grpc.internal.ManagedChannelImpl.<init>(ManagedChannelImpl.java:394)
        at io.grpc.internal.AbstractManagedChannelImplBuilder.build(AbstractManagedChannelImplBuilder.java:350)
        at io.pravega.client.stream.impl.ControllerImpl.<init>(ControllerImpl.java:140)
        at io.pravega.client.stream.impl.ControllerImpl.<init>(ControllerImpl.java:114)
        at io.pravega.client.admin.impl.StreamManagerImpl.<init>(StreamManagerImpl.java:35)
        at io.pravega.client.admin.StreamManager.create(StreamManager.java:28)
        at io.pravega.connectors.flink.utils.SetupUtils.createTestStream(SetupUtils.java:126)
```
**Problem location**
SetupUtils.java

In SetupUtils getControllerUri method:
```    public URI getControllerUri() {
        return URI.create("tcp://" + this.inProcPravegaCluster.getControllerURI());
    }
```
It expects inProcPravegaCluster.getControllerUri to be only <IP:port> value.
However, "tcp://" prefix is already added by inprocPravegaCluster
This causes uri to look like "tcp://tcp:// ..." 

[...]

Issue #42, opened 2017-08-07
Implement batch connector (source only)


EronWright:
**Problem description**
Develop a batch connector to process historical data in a given stream using the Flink Batch API.  A good reason is to be able to use Flink Gelly (graph API) and Flink ML which are based on the batch API.

It is probably sufficient to consider a Pravega stream as a _source_, not as a _sink_, for batch purposes.

**Suggestions for an improvement**
_There's at least two approaches being considered.  One is to treat segments as splits, which seems more likely to leverage data locality and more compatible with fine-grained recovery, but requires client changes.  Another is to synthesize a split per subtask and to use the existing reader group API, which would require fewer changes to the Pravega client but underperform at scale._

Implement an `InputFormat`, which splits an input file or other resource (e.g. Pravega stream) into processable units for Flink to process across a number of subtask instances.  For example, a file-based input format splits the input file(s) into splits representing non-overlapping ranges that correspond to HDFS blocks.    For Pravega, a close analogue would be to treat each segment as an input split.

If location hints are available (on a per-split basis), implement `LocatableInputSplit` and return an instance of `LocatableInputSplitAssigner` from `InputFormat`.

A connector is free to read each split in whatever way it chooses.  It needn't be a file-based access.  For example, if each segment were a split, the connector could use a Pravega Reader API to query a specific segment, e.g. `io.pravega.client.segment.impl.SegmentInputStream` (which the reader group API uses under the hood).

EronWright:
@StephanEwen @fpj In our recent discussion, it was suggested that the existing reader group API be used for batch processing.   Despite its benefits, one concern I have is that it might underperform relative to the segment processing approach, due to the ordering guarantee that the reader group normally provides but isn't (to my knowledge) useful here.    Imagine a stream that has, on average, ten segments covering the keyspace at any given point in time.  There might be hundreds or thousands of segments in the stream as a whole, but scaling the subtasks beyond ten would yield diminishing returns because subtasks would sit idle.  In other words, the scheduling of segments would be overly conservative.

EronWright:
(A visualization to complement my above comment)
![pravega stream viz](https://user-images.githubusercontent.com/1775518/30716032-4f3ff620-9ecd-11e7-8f95-2a3fa378f1b4.png)

fpj:
@EronWright 

[...]

PR #54, opened 2017-09-26
[Issue #42] Add initial batch read support for Flink Pravega Connectors

tzulitai:
**Change log description**
- 2fe0e92 preliminary fix for #53 
- 12a2fe3 introduce `FlinkPravegaInputFormat`

**Purpose of the change**
Add batch support to the connector. Users can add the `FlinkPravegaInputFormat` as a source to Flink batch jobs.

**Known pitfalls / limitations**
- Reader group creation is performed on the client. Apparently, Flink's `InitializeOnMaster` and `FinalizeOnMaster` hooks only currently works for `OutputFormat`s, so we can't use them for the one-time reader group creation / deletion.
- The "end of input" marker is quite brittle at the moment. The PR simply uses a return of `null` on of the last read event `EventRead.getEvent()` as an indicator. We might want to add some retry logic there to strengthen that, and avoid cases where the `null` return was just caused by some timeout caused by temporary read hiccups.

**How to verify it**
Added integration test: `FlinkPravegaInputFormatTest`. The test basically writes some events to Pravega, and uses a batch Flink job to read all written events and verifies that everything is properly read when the batch job finishes.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
@@ -1,0 +1,179 @@
+/**
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.pravega.connectors.flink;
+
+import com.google.common.base.Preconditions;
+
+import io.pravega.client.admin.ReaderGroupManager;
+import io.pravega.client.stream.EventRead;
+import io.pravega.client.stream.EventStreamReader;
+import io.pravega.client.stream.ReaderConfig;
+import io.pravega.client.stream.ReaderGroupConfig;
+
+import org.apache.flink.api.common.io.GenericInputFormat;
+import org.apache.flink.api.common.io.InputFormat;
+import org.apache.flink.core.io.GenericInputSplit;
+import org.apache.flink.streaming.util.serialization.DeserializationSchema;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.Set;
+
+import static io.pravega.connectors.flink.util.FlinkPravegaUtils.createPravegaReader;
+import static io.pravega.connectors.flink.util.FlinkPravegaUtils.generateRandomReaderGroupName;
+
+/**
+ * A Flink {@link InputFormat} that can be added as a source to read from Pravega in a Flink batch job.
+ */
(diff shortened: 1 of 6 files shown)


fpj:
@tzulitai  We will be able to handle better end of input once we implement pravega/pravega#1916. If you want this merged before we complete that work, then I suggest we create an issue to track it and fix it later.

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
```diff
@@ -0,0 +1,179 @@
...
+    @Override
+    public boolean reachedEnd() throws IOException {
+        // look ahead to see if we have reached the end of input
+        try {
+            this.lastReadAheadEvent = pravegaReader.readNextEvent(eventReadTimeout);
+        } catch (Exception e) {
+            throw new IOException("Failed to read next event.", e);
+        }
+
+        // TODO this "end of input" marker is too brittle, as the timeout could easily be a temporary hiccup;
```
fpj: See pravega/pravega#1916. We can fix this here once we complete the work of that issue.

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
```diff
@@ -0,0 +1,179 @@
...
+import static io.pravega.connectors.flink.util.FlinkPravegaUtils.generateRandomReaderGroupName;
+
+/**
+ * A Flink {@link InputFormat} that can be added as a source to read from Pravega in a Flink batch job.
+ */
+public class FlinkPravegaInputFormat<T> extends GenericInputFormat<T> {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final long DEFAULT_EVENT_READ_TIMEOUT = 1000;
```
fpj: This is fine, but on the optimistic side, perhaps a value like 10s would suit better. If this is a batch read and all the stream data is there, then we should in the absence of any glitches never time out.

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
```diff
@@ -0,0 +1,179 @@
...
+        Preconditions.checkNotNull(deserializationSchema, "deserializationSchema");
+
+        this.controllerURI = controllerURI;
+        this.scopeName = scope;
+        this.deserializationSchema = deserializationSchema;
+        this.streamNames = streamNames;
+        this.startTime = startTime;
+        this.readerGroupName = generateRandomReaderGroupName();
+
+        // TODO: This will require the client to have access to the pravega controller and handle any temporary errors.
```
fpj: As a general suggestion, I like the idea of creating issues for TODOs in the code so that we can track what we need to fix in the future.

EronWright:
I am going to merge this now and ensure that any unfinished aspect is tracked in a follow-up issue.

PR #61, opened 2017-10-12
[issue-32] Table Connector

### LINKED ISSUE #32: Implement table connector (streaming only)
**Problem description**
A Flink table connector (`TableSource`/`TableSink`) is another type of connector, needed to support Flink's Table/SQL library.   Two variants exist, one for integration with the streaming API (e.g. `StreamTableSource`), the other for the batch API (e.g. `BatchTableSource`).   The implementation typically builds on the classic streaming or batch connector.

**Implementation Details**
The duty of a table source is to return a `DataStream<Row>` where `Row` is a core data type of Flink Table.  A typical table source takes a `DeserializationSchema<Row>` as a parameter, to support conversion from an arbitrary byte message to a `Row`.  Flink provides `JsonRowDeserializationSchema` and `AvroRowDeserializationSchema` for convenience.

A table sink is similar to a source but has some interesting implementation details, building on the concept of [Streams & Tables](http://s.apache.org/beam-streams-tables).  A basic implementation may extend `AppendStreamTableSink` to indicate that it accepts inserts only.    A fancy sink may extend other variants to support deletes, upserts, etc.

More information here: [Flink documentation](https://ci.apache.org/projects/flink/flink-docs-release-1.3/dev/table/sourceSinks.html)

EronWright:
**Change log description**
- add `FlinkPravegaTableSource`
- add `FlinkPravegaTableSink`
- add `FlinkTableITCase`
- add JSON serialization schema
- add jackson as a 'provided' (by Flink) dependency

**Purpose of the change**
Introduces Flink Table API support, including a table source and append-only table sink.  Closes #32 .

**How to verify it**
Execute integration test `FlinkTableITCase`.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
@@ -1,0 +1,113 @@
+/**
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.pravega.connectors.flink;
+
+import io.pravega.connectors.flink.util.StreamId;
+
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.streaming.util.serialization.DeserializationSchema;
+import org.apache.flink.table.sources.StreamTableSource;
+import org.apache.flink.types.Row;
+
+import java.net.URI;
+import java.util.Collections;
+import java.util.function.Function;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * A table source to produce a streaming table from a Pravega stream.
+ */
+public class FlinkPravegaTableSource implements StreamTableSource<Row> {
+
+    /** The Pravega controller endpoint. */
+    private final URI controllerURI;
+
+    /** The Pravega stream to use. */
+    private final StreamId stream;
+
+    /** The start time from when to read events from. */
+    private final long startTime;
+
+    /** Deserialization schema to use for Pravega stream events. */
+    private final DeserializationSchema<Row> deserializationSchema;
+
+    /** Type information describing the result type. */
+    private final TypeInformation<Row> typeInfo;
+
+    /**
+     * Creates a Pravega {@link StreamTableSource}.
+     *
+     * <p>The {@code deserializationSchemaFactory} supplies a {@link DeserializationSchema}
+     * based on the result type information.
+     *
+     * @param controllerURI                The pravega controller endpoint address.
+     * @param stream                       The stream to read events from.
+     * @param startTime                    The start time from when to read events from.
+     * @param deserializationSchemaFactory The deserialization schema to use for stream events.
+     * @param typeInfo                     The type information describing the result type.
+     */
+    public FlinkPravegaTableSource(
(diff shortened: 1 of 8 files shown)


EronWright:
Note that the `JsonRowDeserializationSchema` and `JsonRowSerializationSchema` classes were copied from the Flink Kafka connector.  I'll follow up with the original authors as to whether they could be moved to the flink-table library.  Unsure as to whether any attribution is needed.

chrisdail:
Providing a JsonSerializationSchema makes sense here. We may want to take the next step and also provide a Pravega `io.pravega.client.stream.Serializer` that works with Json. That way someone can write a standalone app to write the data and read it using the table connector.

Extending that, if we have a `JsonSerializer` for Pravega, perhaps we should use the `PravegaDeserializationSchema` that already exists and use the JsonSerializer in conjunction with that for the table connector.

EronWright:
I think this serializer may be too specialized to be replaced with a
generic Pravega serializer since it is coded for the Row class and not
based on Jackson databinding.

I agree completely that the Pravega client should provide a Json serializer
based on data binding.  We could provide something in the samples.  Will
file.

EronWright:
@tzulitai I invited you as a collaborator, then I can request a review from you.

src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
```diff
@@ -0,0 +1,113 @@
...
+        this.deserializationSchema = deserializationSchemaFactory.apply(typeInfo);
+    }
+
+    /**
+     * NOTE: This method is for internal use only for defining a TableSource.
+     *       Do not use it in Table API programs.
+     */
+    @Override
+    public DataStream<Row> getDataStream(StreamExecutionEnvironment env) {
+        FlinkPravegaReader<Row> kafkaConsumer = createFlinkPravegaReader();
```
vijikarthi: Please change the field name to `pravegaReader`

src/main/java/io/pravega/connectors/flink/serialization/JsonRowDeserializationSchema.java
```diff
@@ -0,0 +1,113 @@
...
+     *
+     * @param typeInfo   Type information describing the result type. The field names are used
+     *                   to parse the JSON file and so are the types.
+     */
+    public JsonRowDeserializationSchema(TypeInformation<Row> typeInfo) {
+        Preconditions.checkNotNull(typeInfo, "Type information");
+        this.typeInfo = typeInfo;
+
+        this.fieldNames = ((RowTypeInfo) typeInfo).getFieldNames();
+        this.fieldTypes = ((RowTypeInfo) typeInfo).getFieldTypes();
```
vijikarthi: Are we expecting `fieldNames` and `fileldTypes` size to match here?

src/main/java/io/pravega/connectors/flink/serialization/JsonRowDeserializationSchema.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Configures the failure behaviour if a JSON field is missing.
+     *
+     * <p>By default, a missing field is ignored and the field is set to null.
+     *
+     * @param failOnMissingField Flag indicating whether to fail or not on a missing field.
+     */
+    public void setFailOnMissingField(boolean failOnMissingField) {
```
vijikarthi: Can this flag be part of the constructor instead?

src/main/java/io/pravega/connectors/flink/serialization/JsonRowDeserializationSchema.java
EronWright: I'm emphasizing keeping this source file in sync with flink-kafka, and so don't plan to improve it.

src/main/java/io/pravega/connectors/flink/serialization/JsonRowDeserializationSchema.java
EronWright: (this would undermine the use of a method reference in the `TableSource` constructor)

Issue #67, opened 2017-11-15
Develop a unit test for FlinkPravegaWriter


EronWright:
**Problem description**
The `FlinkPravegaWriter` has a pair of existing tests, `FlinkPravegaWriterTest` and `FlinkExactlyOncePravegaWriterTest`, that are best understood as integration tests.   Overall coverage is low.  

**Suggestions for an improvement**
- Develop a unit test that uses a mock Pravega event writer to fully exercise the `FlinkPravegaWriter`.
- Consolidate the existing tests into a test called `FlinkPravegaWriterITCase`.

PR #72, opened 2017-11-17
[issue-70] Fix accounting of pending writes in FlinkPravegaWriter

### LINKED ISSUE #70: FlinkPravegaWriter cannot be flushed following a write error
**Problem description**
When `close` or `flush` is called after a write that has failed internally, the method will stall forever.

**Problem location**
The problem is due to bad accounting of pending writes based on `pendingWritesCount`.  The `flush`  method waits for the count to reach zero, but the count is not properly decremented on error.  ([ref](https://github.com/pravega/flink-connectors/blob/r0.1/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java#L162))

Decrement the write count irrespective of success or failure.

EronWright:
**Change log description**
- Fixed accounting of pending writes in FlinkPravegaWriter.

**Purpose of the change**
Prevent a stall in the sink following a write error.  Closes #70.   

**What the code does**

**How to verify it**
- use the IT test: `FlinkPravegaWriterTest`.
- note that #67 will provide additional verification tests.


EronWright:
I'll reopen this once #30 is merged.

Issue #56, opened 2017-10-10
Batch connector doesn't rewind on task failover


EronWright:
**Problem description**
Flink supports a number of failover strategies when a task fails, by default a global restart of the job.   None of the restart strategies work correctly with `FlinkPravegaInputFormat` because it doesn't rewind the reader group to the start position.

**Suggestions for an improvement**
- Discard the reader group state upon a global restart.  
- Validate that the job is configured to use the global failover strategy.

**Additional Background**
The restart procedure for an input format normally works as follows:
1. `InputFormat::new` is called by the client, then serialized into a job graph.
2. `InputFormat::configure` is called once by the JM when the execution graph is built.
3. `InputFormat::createInputSplits` is called once by the JM when the execution graph is built.
4. `InputFormat::getInputSplitAssigner(splits)` is called for each execution (incl. `resetForNewExecution`)
5. `InputSplitAssigner::getNextInputSplit` is called by the JM to produce a split for processing by an idle task.

In words, the 'assigner' is the iterator and is reset by the 'global' failover strategy using `resetForNewExecution`.

fpj:
As I understand the description, this issue is about implementing the missing mechanism in the connector, not about a missing feature in Pravega, is it right?

The Pravega reader group has the notion of checkpoints, which can be used to rewind to a previous position.

tzulitai:
@EronWright I'll apply a PR for this issue.

tzulitai:
Hi @EronWright! I've proposed a fix for this issue: https://github.com/pravega/flink-connectors/pull/77

EronWright:
It is becoming clear that the true fix is to use the Pravega batch API - see #65

tzulitai:
This is now also fixed after 5eea41a74c543f859e0ab12894fb7122c4f2bb99.
Closing this.

Issue #65, opened 2017-11-02
Use Pravega batch read API with batch connector


fpj:
**Problem description**
We have merged an experimental API for batch reads of a stream:

https://github.com/pravega/pravega/commit/493a9f45c561ec2a07b438f02796f9070db5cb77

The general idea is that a job has access to all segments in parallel, assuming that for batch jobs, we don't care about the order of events in a stream.

The idea is to incorporate this API in the preliminary batch connector developed  in PR #54.

**Problem location**
Batch connectors.

**Suggestions for an improvement**
Use the experimental batch read API

EronWright:
@tzulitai thanks again for picking this up.  I think it is quite urgent, because we're finding that the current implementation is not workable.   As we saw with #77, the reader group imposes a single pass over the data, which is (I believe) a violation of the contract of input format.

Please ensure that the solution is compatible with Flink iteration and the like.  For example, the following should work:
```
DataSource<Integer> source = env.createInput(new FlinkPravegaInputFormat<>(...));
Assert.assertEquals("count is incorrect (first pass)", expectedCount, source.count());
Assert.assertEquals("count is incorrect (second pass)", expectedCount, source.count());
```

PR #77, opened 2017-11-30
[issue-56] Reset reader group on every execution attempt in FlinkPravegaInputFormat

tzulitai:
**Change log description**

This PR addresses #56.

- Recreate a new Pravega reader group on every execution attempt. This is done in the `openInputFormat` method, which is called on every execution attempt (i.e., including failovers).
- Add a new batch read integration test case, which verifies all records are still read if the batch read fails and the job is restarted.

**How to verify it**

The fix is verified with the new test `FlinkPravegaInputFormatTest#testBatchReadWithFailure()`.
That test fails without this fix.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
@@ -15,6 +15,7 @@ import com.google.common.base.Preconditions;
 import io.pravega.client.admin.ReaderGroupManager;
 import io.pravega.client.stream.EventRead;
 import io.pravega.client.stream.EventStreamReader;
+import io.pravega.client.stream.InvalidStreamException;
 import io.pravega.client.stream.ReaderConfig;
 import io.pravega.client.stream.ReaderGroupConfig;
 
@@ -48,8 +49,9 @@ public class FlinkPravegaInputFormat<T> extends GenericInputFormat<T> {
     // The scope name of the destination stream.
     private final String scopeName;
 
-    // The readergroup name to coordinate the parallel readers. This should be unique for a Flink job.
-    private final String readerGroupName;
+    // The prefix for Pravega reader group names to coordinate the parallel readers. This should be unique for a Flink job.
+    // The actual reader group name is recreated for each execution attempt of the job.
+    private final String readerGroupNamePrefix;
 
     // The names of Pravega streams to read
     private final Set<String> streamNames;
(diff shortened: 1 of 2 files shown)


src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
```diff
@@ -176,4 +184,36 @@ public T nextRecord(T t) throws IOException {
...
+
+    private void setupReaderGroupForCurrentExecution() {
+        try (ReaderGroupManager readerGroupManager = ReaderGroupManager.withScope(scopeName, controllerURI)) {
+            final int currentAttemptNum = getRuntimeContext().getAttemptNumber();
+
+            // delete reader group from last execution attempt, if any
+            if (currentAttemptNum > 0) {
+                try {
+                    readerGroupManager.deleteReaderGroup(getReaderGroupNameForExecutionAttempt(currentAttemptNum - 1));
+                } catch (RuntimeException e) {
```
tzulitai: This part is certainly not that nice.
The main issue is that there currently is no "global failure hook" for input formats when an execution attempt fails, so there is no simple way to delete reader groups only once on failure.

tzulitai: I may also be doing this overly complicated.
Is there a better approach to reset the reader group state, instead of creating a new reader group and deleting the previous one?

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
```diff
@@ -176,4 +184,36 @@ public T nextRecord(T t) throws IOException {
...
+                    // ignore InvalidStreamExceptions, which is the result of duplicate reader group deletions
+                    // (the reader group deletion operation is not idempotent); rethrow all other exceptions
```
tzulitai: This part is certainly not that nice.
The main issue is that there currently is no "global failure hook" for input formats when an execution attempt fails, so there is no simple way to delete reader groups only once on failure.

tzulitai: I may also be doing this overly complicated.
Is there a better approach to reset the reader group state, instead of creating a new reader group and deleting the previous one?

EronWright:
Given the issues you mention, plus the possibility of fine-grained recovery in Flink, the reader group API is an especially poor fit for the batch scenario.   The Pravega client actually has a new batch client API that is much better suited here.  See [hadoop-connectors](https://github.com/pravega/hadoop-connectors). 

Let's close this PR and focus on using the batch client API, OK?

tzulitai:
@EronWright Sure, that makes sense. Will close this and pick up #65.

PR #80, opened 2017-12-01
Use lang3.RandomStringUtils

jdmaguire:
I think we accidentally used lang(2)'s RandomStringUtils. This was being silently fetched into the uber jar because Pravega had it as a dependency. They've dropped it in on master.

This class is deprecated. Talking it over with Eron, we feel using a deprecated class is better than bringing the entire commons-text package.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,6 +69,8 @@ dependencies {
     compile group: 'io.pravega', name: 'pravega-shared-protocol', version: pravegaVersion
     compile group: 'io.pravega', name: 'pravega-shared-controller-api', version: pravegaVersion
     compile group: 'org.slf4j', name: 'slf4j-api', version: slf4jApiVersion
+    compile group: 'org.apache.commons', name: 'commons-lang3', version: apacheCommonsVersion
+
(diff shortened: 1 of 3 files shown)


fpj:
We have gone through the same issue in Pravega, you may want to check the discussion here: https://github.com/pravega/pravega/issues/2064

jdmaguire:
Thanks for the insight @fpj. The trouble at present is this:

- One class was using org.apache.commons.lang.RandomStringUtils. (Perhaps by accident since the tests use lang3's RandomStringUtils)
- Commons-lang _was_ a transitive, compile dependency but no longer is (commons-lang3 3.6 still is)
- This project uses the `-Werror` Gradle flag (therefore can't use deprecated classes without suppressing the warning)

Because of all this, using this project against recent Pravega master throws a ClassDefNotFound. I talked to Eron off-channel about this. He thought suppressing the deprecated warning was the best of many solutions available; I wasn't aware 3.7 removed the deprecation of this class.

fpj:
@jdmaguire @EronWright sounds good.

PR #81, opened 2017-12-05
Add netty and netty-all to dependency list

jdmaguire:
Similar to #80. A dependency is using a class from netty or netty-all that we're not including in the uber jar. This is causing a ClassDefNotFound. (Note: in shadowjars we need to include transitive dependencies explicitly. Both netty and netty-all are transitive dependencies.)


EronWright:
@jdmaguire what class was missing?

Issue #78, opened 2017-11-30
Develop a unit test for FlinkPravegaInputFormat


tzulitai:
**Problem description**

Likewise to #67 and #68, the current tests for the `FlinkPravegaInputFormat` are strictly speaking integration tests that do not have sufficient coverage. We should develop proper unit tests for the connector.

**Suggestions for an improvement**

- Rename `FlinkPravegaInputFormatTest` to `FlinkPravegaInputFormatITCase`.
- Develop a unit test that uses a mock Pravega `EventStreamReader`.

PR #90, opened 2017-12-13
[issue-65] Rework FlinkPravegaInputFormat to use batch client

### LINKED ISSUE #65: Use Pravega batch read API with batch connector

tzulitai:
**Change log description**

- Reworks the `FlinkPravegaInputFormat` to use the experimental Pravega batch client.
- Expand current IT tests for the input format. The tests now additionally cover the following verifications: 1) reading multiple streams with batch, 2) multiple passes, and 3) rewinds input format in case of failures.
- Previous configuration of the input format such as `startReadTimestamp` and `eventReadTimeout` is removed, since they are no longer relevant / useful after switching to the batch client.

**Purpose of the change**

This PR closes #65 and #56.

As discussed in #65 and #56, the use of reader groups / the `EventStreamReader` API does not work well in batch scenarios, which only allows Pravega segments to be read sequentially. There's also the problem that residual reader group states does not (gracefully) allow the input format to have multiple passes, or even re-executions on job failures.

This PR reworks the input format to use the experimental batch client instead.

**How to verify it**

- The original `FlinkPravegaInputFormatTest.testBatchInput()` is expanded to verify multiple passes of the input format, as well as reading a set of multiple input streams.
- A new `FlinkPravegaInputFormatTest.testBatchInputWithFailure()` is added to verify the behaviour in the presence of failures.

The above tests all fail without the rework in `FlinkPravegaInputFormat`.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
@@ -12,32 +12,38 @@ package io.pravega.connectors.flink;
 
 import com.google.common.base.Preconditions;
 
-import io.pravega.client.admin.ReaderGroupManager;
+import io.pravega.client.ClientFactory;
+import io.pravega.client.batch.BatchClient;
+import io.pravega.client.batch.SegmentInfo;
+import io.pravega.client.batch.SegmentIterator;
+import io.pravega.client.segment.impl.Segment;
 import io.pravega.client.stream.EventRead;
-import io.pravega.client.stream.EventStreamReader;
-import io.pravega.client.stream.ReaderConfig;
-import io.pravega.client.stream.ReaderGroupConfig;
+import io.pravega.client.stream.Serializer;
+import io.pravega.client.stream.impl.StreamImpl;
+import io.pravega.connectors.flink.serialization.WrappingSerializer;
+import io.pravega.connectors.flink.util.FlinkPravegaUtils;
 
-import org.apache.flink.api.common.io.GenericInputFormat;
+import org.apache.flink.api.common.io.DefaultInputSplitAssigner;
 import org.apache.flink.api.common.io.InputFormat;
-import org.apache.flink.core.io.GenericInputSplit;
+import org.apache.flink.api.common.io.RichInputFormat;
+import org.apache.flink.api.common.io.statistics.BaseStatistics;
+import org.apache.flink.configuration.Configuration;
 import org.apache.flink.api.common.serialization.DeserializationSchema;
+import org.apache.flink.core.io.InputSplitAssigner;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
 import java.util.Set;
-
-import static io.pravega.connectors.flink.util.FlinkPravegaUtils.createPravegaReader;
-import static io.pravega.connectors.flink.util.FlinkPravegaUtils.generateRandomReaderGroupName;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
```diff
@@ -76,104 +76,112 @@
...
      * @param streamNames           The list of stream names to read events from.
-     * @param startTime             The start time from when to read events from.
-     *                              Use 0 to read all stream events from the beginning.
      * @param deserializationSchema The implementation to deserialize events from pravega streams.
      */
     public FlinkPravegaInputFormat(
             final URI controllerURI,
             final String scope,
             final Set<String> streamNames,
-            final long startTime,
```
EronWright: This is a breaking change (that I am fine with).

src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java
```diff
@@ -0,0 +1,62 @@
...
+package io.pravega.connectors.flink;
+
+import com.google.common.base.Preconditions;
+import io.pravega.client.segment.impl.Segment;
+import org.apache.flink.core.io.InputSplit;
+
+/**
+ * A {@link PravegaInputSplit} corresponds to a Pravega {@link Segment}.
+ */
+public class PravegaInputSplit implements InputSplit {
```
EronWright: Is there any requirement that the split implement equality or comparison checks?

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
tzulitai: Yes. I also think it is ok, since the batch connector is not yet in any official release, correct?

src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java
tzulitai: I don't think so. Though, it would not hurt to implement that (pretty sure we'll need it eventually for the unit tests anyway). Will add.

tzulitai:
@EronWright thanks for the review!
I've addressed your comments. Also, a follow-up issue for writing proper unit tests for the input format was already opened here: https://github.com/pravega/flink-connectors/issues/78.

src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java
```diff
@@ -0,0 +1,104 @@
...
+        return endOffset;
+    }
+
+    // --------------------------------------------------------------------
+
+    @Override
+    public int hashCode() {
+        int result = splitId;
+
+        // Pravega's Segment does not have hashCode implemented
```
EronWright: I think it does, thanks to the `@Data` lombok annotation.

EronWright:
This is great, thanks so much.    If you have a little more time, just make use of `Segment::hashCode()` and we're good to merge.

src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java
tzulitai: Ah, I didn't realize that. Thanks, will change.

tzulitai:
@EronWright updated 👍

Issue #88, opened 2017-12-08
Treat Pravega as a source dependency


EronWright:
**Problem description**
Fix the travis build of the connector to build Pravega automatically, so that the connector may be unit tested by travis.

**Suggestions for an improvement**
Consider using a submodule plus a [composite build](https://docs.gradle.org/current/userguide/composite_builds.html) to incorporate the Pravega build into the connector build.

fpj:
@EronWright I have a few of questions about this issue:
- The travis build is failing because it is looking for 0.3.0-SNAPSHOT, so I think you are trying to address that. Is it necessary that connectors master relies on pravega master, or should we make the connectors depend on the latest released? I can't remember whether we have come to a conclusion about that.
- Would it be optional to build Pravega from source with the changes in PR #98?
- How would it work with releases? For a release of the connectors, it would be good to make sure they depend on a specific version of Pravega.

EronWright:
@fpj I agree there's a big decision point here about how closely coupled the connector should be.   

I made it so that the `pravegaVersion` variable determines the behavior; release versions of Pravega are pulled from Maven while snapshots are built from the submodule.  

Honestly I don't know how this approach will pan out, an in particular how it will cooperate with shading of Pravega.   At the moment it seems to work nicely.

PR #93, opened 2017-12-20
[Issue 92] Fixed SegmentInfo API call change and checkstyle errors

### LINKED ISSUE #92: Fix batch connector to support Pravega segment truncation
**Problem description**
The SegmentInfo API has changed with this new commit from Pravega which is causing the batch connector code compilation to fail. Fix the API call accordingly.
https://github.com/pravega/pravega/commit/be358929db2f3e1b52f408b155ed2e37d88f910b

**Problem location**
https://github.com/pravega/flink-connectors/blob/master/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java#L140

vijikarthi:
**Change log description**
- Update to Pravega-0.2 (rc0)
- Incorporate changes to Pravega client related to segment truncation.
- Fix checkstyle failures.

**Purpose of the change**
Ensure compatibility with Pravega 0.2.  Closes #92.

**How to verify it**
Ran integration tests and confirmed

### DIFF
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -71,11 +71,9 @@
     </module>
     <module name="PackageName"/>
     <module name="ParameterName"/>
-    <!--
     <module name="StaticVariableName">
         <property name="format" value="Static non-final variables are not allowed."/>
     </module>
-    -->
     <module name="TypeName"/>
     <module name="AvoidStarImport">
        <!--    <property name="excludes" value="java.io,java.net,java.lang.Math"/> -->
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -71,9 +71,11 @@
     </module>
     <module name="PackageName"/>
     <module name="ParameterName"/>
+    <!--
     <module name="StaticVariableName">
         <property name="format" value="Static non-final variables are not allowed."/>
     </module>
+    -->
     <module name="TypeName"/>
     <module name="AvoidStarImport">
        <!--    <property name="excludes" value="java.io,java.net,java.lang.Math"/> -->
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,7 +15,7 @@ mockitoVersion=1.10.19
 
 # Version and base tags can be overridden at build time.
 connectorVersion=0.2.0-SNAPSHOT
-pravegaVersion=0.3.0-SNAPSHOT
+pravegaVersion=0.2.0
 apacheCommonsVersion=3.7
 
 # These properties are only needed for publishing to maven central
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,7 +15,7 @@ mockitoVersion=1.10.19
 
 # Version and base tags can be overridden at build time.
 connectorVersion=0.2.0-SNAPSHOT
-pravegaVersion=0.2.0-SNAPSHOT
+pravegaVersion=0.3.0-SNAPSHOT
 apacheCommonsVersion=3.7
 
 # These properties are only needed for publishing to maven central
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
@@ -137,7 +137,7 @@ public class FlinkPravegaInputFormat<T> extends RichInputFormat<T, PravegaInputS
                 for (Iterator<SegmentInfo> segmentInfos = batchClient.listSegments(new StreamImpl(scopeName, stream)); segmentInfos.hasNext(); ) {
                     SegmentInfo segmentInfo = segmentInfos.next();
                     Segment segment = segmentInfo.getSegment();
-                    splits.add(new PravegaInputSplit(splits.size(), segment, 0, segmentInfo.getWriteOffset()));
+                    splits.add(new PravegaInputSplit(splits.size(), segment, segmentInfo.getStartingOffset(), segmentInfo.getWriteOffset() - 1));
                 }
             }
         }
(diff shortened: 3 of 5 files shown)


checkstyle/checkstyle.xml
```diff
@@ -71,9 +71,11 @@
     </module>
     <module name="PackageName"/>
     <module name="ParameterName"/>
+    <!--
```
vijikarthi: I have commented this *temporarily* since a non-final static reference was used in `FlinkPravegaInputFormatITCase` code.

gradle.properties
```diff
@@ -15,7 +15,7 @@ mockitoVersion=1.10.19
 
 # Version and base tags can be overridden at build time.
 connectorVersion=0.2.0-SNAPSHOT
-pravegaVersion=0.2.0-SNAPSHOT
+pravegaVersion=0.3.0-SNAPSHOT
```
vijikarthi: Pravega `master` is now in 0.3.0-SNAPSHOT. Not sure if we need to port the changes to 0.2.0-SNAPSHOT branch as well?

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
```diff
@@ -137,7 +137,7 @@ public BaseStatistics getStatistics(BaseStatistics cachedStatistics) throws IOEx
                 for (Iterator<SegmentInfo> segmentInfos = batchClient.listSegments(new StreamImpl(scopeName, stream)); segmentInfos.hasNext(); ) {
                     SegmentInfo segmentInfo = segmentInfos.next();
                     Segment segment = segmentInfo.getSegment();
-                    splits.add(new PravegaInputSplit(splits.size(), segment, 0, segmentInfo.getLength()));
+                    splits.add(new PravegaInputSplit(splits.size(), segment, 0, segmentInfo.getWriteOffset()));
```
EronWright: The constructor of `PravegaInputSplit` is `(..., startOffset, endOffset)`, and the range is inclusive, so this should read:
```
splits.add(new PravegaInputSplit(splits.size(), segment, segmentInfo.getReadOffset(), segmentInfo.getWriteOffset() - 1));
```

@tkaitchuck you agree?

checkstyle/checkstyle.xml
EronWright: Use a targeted suppression if possible.
https://stackoverflow.com/questions/27688426/ignoring-of-checkstyle-warnings-with-annotation-suppresswarnings

gradle.properties
EronWright: I think you're referring to the connector's `release-nautilus-0.3` branch.   These changes are not needed there.

vijikarthi: No. I am referring to current `master` from which we may have to cut a release branch `release-nautilus-0.4` for Pravega 0.2.0 release support.

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
```diff
@@ -137,7 +137,7 @@ public BaseStatistics getStatistics(BaseStatistics cachedStatistics) throws IOEx
                 for (Iterator<SegmentInfo> segmentInfos = batchClient.listSegments(new StreamImpl(scopeName, stream)); segmentInfos.hasNext(); ) {
                     SegmentInfo segmentInfo = segmentInfos.next();
                     Segment segment = segmentInfo.getSegment();
-                    splits.add(new PravegaInputSplit(splits.size(), segment, 0, segmentInfo.getLength()));
+                    splits.add(new PravegaInputSplit(splits.size(), segment, segmentInfo.getStartingOffset(), segmentInfo.getWriteOffset() - 1));
```
EronWright: Turns out that the call to `batchClient::readSegment` has an exclusive `endingOffset`.   But the `PravegaInputSplit` is written to be inclusive.  So we need to change the call to `readSegment` below, or change the `PravegaInputSplit`.

cc @tkaitchuck

tkaitchuck: There should be no -1 here

fpj: @EronWright I'm not sure I follow why do you say that `PravegaInputSplit` is written to be inclusive. There is no description mentioning it, although there is this precondition, which is perhaps what is making you say it:

https://github.com/pravega/flink-connectors/blob/master/src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java#L36

If there is any reason for it to be inclusive, then I think we do need to subtract and add, but it is not clear that it is needed. Could you clarify, please?

tzulitai: There isn't any specific requirement for the `PravegaInputSplit` to have an inclusive end offset.
Though, it would make sense to clarify this in the Javadocs.

tzulitai:
A separate question: is there any consideration to include building Pravega as part of the connector build, so that this can be avoided in the future?

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
tkaitchuck: Sorry for the confusion. I am removing this parameter to avoid this in the future: https://github.com/pravega/pravega/pull/2216

EronWright:
@tzulitai yes, I opened #88 recently to consider building Pravega when we build the connector.

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
EronWright: I agree that the split could easily have an exclusive `endOffset`, and that its javadocs should be updated in any case.   My only objection is that the name `endOffset` suggests inclusiveness (to me).   I can get over that.

PR #108, opened 2018-03-21
[issue-74] added scope/stream context to checkpoint state for recovering right context

### LINKED ISSUE #74: Recovered commit should apply to original stream
**Problem description**
The connector is designed to handle an edge case where the transaction associated with a given savepoint remains uncommitted due to a crash before `notifyCheckpointComplete`.   In that situation, a subsequent job based on that savepoint recovers the transaction during `restoreState`.    The subject of this bug is that the new job might point to a different stream from that of the original job.  The code in `recoverState` should use the old stream name when looking for transactions to recover.

**Problem location**
`FlinkPravegaWriter`

Record the old name as part of the tuple that is returned in `savepoint`.

vijikarthi:
**Change log description**
Modified checkpoint state for the `FlinkPravegaWriter` to include scope and stream along with the transaction id.

**Purpose of the change**
To recover right context/state information during `restoreState` API call.  Closes #74 .

**What the code does**
Earlier only transaction ID was stored as checkpoint state and now the state information will have both scope and stream name along with the transaction id.

**How to verify it**
Running the `FlinkPravegaWriterITCase->testExactlyOnceWriter` should trigger restoreState and the writer instance should be created using stream/scope information from the state object and will be used to commit any pending transactions.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
@@ -33,6 +33,7 @@ import java.nio.ByteBuffer;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
@@ -40,6 +41,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.groupingBy;
 
 /**
  * Flink sink implementation for writing into pravega storage.
@@ -49,7 +52,7 @@ import java.util.stream.Collectors;
 @Slf4j
 public class FlinkPravegaWriter<T>
         extends RichSinkFunction<T>
-        implements ListCheckpointed<FlinkPravegaWriter.PravegaState>, CheckpointListener {
+        implements ListCheckpointed<FlinkPravegaWriter.PendingTransaction>, CheckpointListener {
 
     private static final long serialVersionUID = 1L;
 
@@ -215,7 +218,7 @@ public class FlinkPravegaWriter<T>
     // ------------------------------------------------------------------------
 
     @Override
-    public List<PravegaState> snapshotState(long checkpointId, long checkpointTime) throws Exception {
+    public List<PendingTransaction> snapshotState(long checkpointId, long checkpointTime) throws Exception {
         return writer.snapshotState(checkpointId, checkpointTime);
     }
 
@@ -226,9 +229,9 @@ public class FlinkPravegaWriter<T>
      * Note: {@code restoreState} is called before {@code open}.
      */
     @Override
-    public void restoreState(List<PravegaState> list) throws Exception {
+    public void restoreState(List<PendingTransaction> pendingTransactionList) throws Exception {
         initializeInternalWriter();
-        writer.restoreState(list);
+        writer.restoreState(pendingTransactionList);
     }
 
     /**
@@ -371,9 +374,9 @@ public class FlinkPravegaWriter<T>
             pravegaWriter.close();
         }
 
-        abstract List<PravegaState> snapshotState(long checkpointId, long checkpointTime) throws Exception;
+        abstract List<PendingTransaction> snapshotState(long checkpointId, long checkpointTime) throws Exception;
 
-        abstract void restoreState(List<PravegaState> list) throws Exception;
+        abstract void restoreState(List<PendingTransaction> pendingTransactionList) throws Exception;
 
         abstract void notifyCheckpointComplete(long checkpointId) throws Exception;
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -531,31 +532,73 @@ public void restoreState(List<UUID> list) throws Exception {
...
-                            log.debug("{} - committed checkpoint transaction {}", name(), txnId);
-
-                        } else if (status == Transaction.Status.COMMITTED || status == Transaction.Status.COMMITTING) {
-                            // that the common case
-                            log.debug("{} - at restore, transaction {} was already committed", name(), txnId);
+            Exception exception = null;
+            EventStreamWriter<T> restorePravegaWriter = null;
+            ClientFactory restoreClientFactory = null;
+            PravegaState pravegaState = list != null && list.size() > 0 ? list.get(0) : null;
+            if (pravegaState != null) {
```
EronWright: I believe we must process all states in the list, not just the first item in the list.    This implies the numerous writers must be opened.  Java 8 streams has a nice `groupingBy` construct to make it easy to gather all the txns for a given scope/stream pair ([ref](http://www.baeldung.com/java-groupingby-collector)).

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -682,4 +725,31 @@ private void checkWriteError() throws Exception {
         }
     }
 
+    /*
+     * Pravega state snapshot representing combinations of transaction id, scope and stream
+     */
+    static class PravegaState implements Serializable {
```
EronWright: A better name for this class would be `PendingTransaction`

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
StephanEwen: That's right. You may have more than one state here in a scale-down case, where operator gets the state from multiple previous operators.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -531,33 +539,103 @@ public void restoreState(List<UUID> list) throws Exception {
...
+            }
+
+            Exception exception = null;
+
+            EventStreamWriter<T> restorePravegaWriter = null;
+
+            ClientFactory restoreClientFactory = null;
+
+            Map<String, Map<String, List<PendingTransaction>>> pendingTransactionsMap =
+                    pendingTransactionList.stream().collect(groupingBy(PendingTransaction::getScope, groupingBy(PendingTransaction::getStream)));
```
EronWright: Sorry to ask this, but could you rewrite it such that you group on the composite key of scope + stream?  That would eliminate the nested loop.  I think you can simply do:
```java
import io.pravega.client.stream.Stream;

Map<Stream, List<PendingTransaction>> pendingTransactionsMap = pendingTransactionList
  .stream()
  .collect(groupingBy(t -> Stream.of(t.getScope(), t.getStream());

```

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -531,33 +539,103 @@ public void restoreState(List<UUID> list) throws Exception {
...
+
+            if (pendingTransactionList == null || pendingTransactionList.size() == 0) {
+                return;
+            }
+
+            Exception exception = null;
+
+            EventStreamWriter<T> restorePravegaWriter = null;
+
+            ClientFactory restoreClientFactory = null;
```
EronWright: I bet you can eliminate these local variable declarations using try-with-resources.
```
try(
  ClientFactory restoreClientFactory = ..., 
  EventStreamWriter<T> restorePravegaWriter = ...
) {
 ...
}
```

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -531,33 +539,103 @@ public void restoreState(List<UUID> list) throws Exception {
...
+                        restorePravegaWriter = restoreClientFactory.createEventWriter(stream,
+                                eventSerializer,
+                                writerConfig);
+                        log.info("restore state for the scope: {} and stream: {}", scope, stream);
+
+                        List<PendingTransaction> pendingTransactions = transactionsEntry.getValue();
+
+                        for (PendingTransaction pendingTransaction : pendingTransactions) {
+                            UUID txnId = pendingTransaction.getUuid();
+                            if (txnId != null) {
```
EronWright: Please don't check for null here, instead add a precondition to the `PendingTransaction` constructor that validates the `txnId`.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -682,4 +760,31 @@ private void checkWriteError() throws Exception {
...
+        private final String scope;
+        private final String stream;
+
+        public PendingTransaction(UUID uuid, String scope, String stream) {
+            this.uuid = uuid;
+            this.scope = scope;
+            this.stream = stream;
+        }
+
+        public UUID getUuid() {
```
EronWright: Ideally this property would be named `txnId`

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
vijikarthi: Yes, I have started with that approach but changed it since the `StreamImpl` class does not implement `equals()` and `hashcode()` methods which are necessary to uniquely identify the map key.

EronWright: It actually does thanks to the `@Data` annotation.  Personally I don't like Lombok for the reason that it is too subtle.

PR #109, opened 2018-03-21
[issue-25] added logback configuration for tests to write to both file and stdout

### LINKED ISSUE #25: Add logging support to tests
**Problem description**
All tests are currently timing out in my setup. I went to build looking for logs and didn't find any. It would be useful to have slf4j logging dumped by default to files so that we can inspect in the case of test failures.

**Problem location**
Tests, logging.

Add slf4j logging and configuration.

vijikarthi:
**Change log description**
Added file based logging option for the connector test module

**Purpose of the change**
To support file based logging in addition to the stdout logging. Closes #25 .

**What the code does**
when the connector test module is running, the test logs are recorded in `connector-test.log` file under the root project location.

**How to verify it**
run `./gradlew clean build` and it will generate `connector-test.log` file under the root project location which will have the test outputs.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,6 +65,13 @@ configurations {
     testCompile.exclude group: 'log4j', module: 'log4j'
 }
 
+test {
+    testLogging {
+        events = ["passed", "failed", "skipped"]
+        showStandardStreams = false
+    }
+}
+
 dependencies {
     compile group: 'io.pravega', name: 'pravega-client', version: pravegaVersion
     compile group: 'io.pravega', name: 'pravega-common', version: pravegaVersion
--- a/build.gradle
+++ b/build.gradle
@@ -67,8 +67,13 @@ configurations {
 
 test {
     testLogging {
-        events = ["passed", "failed", "skipped"]
-        showStandardStreams = false
+        onOutput { descriptor, event ->
+            if (project.hasProperty('logOutput')) {
+                events = ["passed", "failed", "skipped"]
+                showStandardStreams = true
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -65,6 +65,13 @@ configurations {
     testCompile.exclude group: 'log4j', module: 'log4j'
 }
 
+test {
+    testLogging {
+        events = ["passed", "failed", "skipped"]
+        showStandardStreams = false
```
vijikarthi: We can change this flag to true to see the `stdout` results on console but the results are verbose and hence turned off by default.

EronWright:
Considering that junit produces a report file that nicely integrates the stdout, what is the merit of writing to a log file?   Is the verbosity different between the console and the file?   Maybe we could enable full stdout logging in travis.

build.gradle
```diff
@@ -67,8 +67,13 @@ configurations {
 
 test {
     testLogging {
-        events = ["passed", "failed", "skipped"]
-        showStandardStreams = false
+        onOutput { descriptor, event ->
+            if (project.hasProperty('logOutput')) {
```
vijikarthi: This change along with the fix (https://github.com/pravega/flink-connectors/commit/c872b3c0d7052b4f815275770145ccc7bbcffd30) to travis config will enable logging when Travis builds the project. On standalone mode, we need to pass the `-PlogOutput` property to the gradle command if we want to see the build/test output in console `./gradlew test -PlogOutput`

PR #114, opened 2018-03-29
[issue-111] added environment variable to enable console logging

vijikarthi:
**Change log description**
Added environment variable to enable console logging in Travis

**Purpose of the change**
To log the connector build/test execution results


vijikarthi:
The log file size is exceeding 4 MB limit of Travis. I am thinking of changing the log level to `ERROR` from `INFO`. I wanted to know if there any other options we have other than changing the level?

Issue #20, opened 2017-05-09
Remove unneeded dependencies


EronWright:
**Problem description**
Some of the 'shared-*' dependencies might not be needed, e.g. shared-metrics.  

**Problem location**
build.gradle

**Suggestions for an improvement**

vijikarthi:
closed https://github.com/pravega/flink-connectors/pull/116

PR #116, opened 2018-04-03
[Issue-112] update Pravega sub-module to latest master commit 

### LINKED ISSUE #112: update pravega submodule to the latest master commit
**Problem description**
The pravega sub-module in the connector project is pointing to a old commit which needs to be refreshed to latest master commit. Fix any issues that may arise due to any changes to Pravega client code (streamcuts API) .

vijikarthi:
**Change log description**
- updated Pravega sub-module dependency to the latest master commit version of Pravega. 
- updated gradle version to match Pravega's gradle version i.e., 4.5.
- updated Flink reader and batch connector implementation to resolve breaking API changes (`StreamCuts` reader API changes) 
- updated dependency versions to match the Pravega client (jackson, lombok, slf4j)
- forced certain dependencies to be compile-only (findbugs, jsr305, lombok) or provided (slf4j)
- forced `io.netty:netty-tcnative-boringssl-static:2.0.8.Final` due to pravega/pravega#2464
- fixed shading of `netty-tcnative-boringssl-static`
- generated proper 'provided' dependencies in published POM via a custom `shadowOnly` configuration

**Purpose of the change**
- To support Flink connector to with latest Pravega/master commit
- Closes #112 and #20 

**How to verify it**
`./gradlew clean build` - all system/unit test cases to pass


src/main/java/io/pravega/connectors/flink/ReaderCheckpointHook.java
```diff
@@ -103,7 +100,7 @@ public void restoreCheckpoint(long checkpointId, Checkpoint checkpoint) throws E
         // checkpoint can be null when restoring from a savepoint that
         // did not include any state for that particular reader name
         if (checkpoint != null) {
-            this.readerGroup.resetReadersToCheckpoint(checkpoint);
+            this.readerGroup.resetReaderGroup(ReaderGroupConfig.builder().startFromCheckpoint(checkpoint).build());
```
EronWright: I know that `resetReadersToCheckpoint` was marked deprecated but I think we should challenge that notion, and continue to use it for now.   Notice that the 'new' approach (as shown here) discards the earlier reader group configuration.  For example, a non-default value for `automaticCheckpointIntervalMillis` would be lost.

src/test/java/io/pravega/connectors/flink/utils/SetupUtils.java
```diff
@@ -171,8 +170,7 @@ public void createTestStream(final String streamName, final int numSegments)
         final String readerGroup = "testReaderGroup" + this.scope + streamName;
         readerGroupManager.createReaderGroup(
                 readerGroup,
-                ReaderGroupConfig.builder().startingTime(0).build(),
-                Collections.singleton(streamName));
+                ReaderGroupConfig.builder().stream(this.scope + "/" + streamName).build());
```
EronWright: Please use the typed variant, e.g.
```
ReaderGroupConfig.builder()
  .stream(Stream.of(this.scope, streamName))
  .build();
```

src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java
```diff
@@ -134,11 +137,14 @@ public BaseStatistics getStatistics(BaseStatistics cachedStatistics) throws IOEx
...
-                    Segment segment = segmentInfo.getSegment();
-                    // segmentInfo.getWriteOffset() will give the length of the segment.
-                    splits.add(new PravegaInputSplit(splits.size(), segment, segmentInfo.getStartingOffset(), segmentInfo.getWriteOffset()));
+
+                Iterator<SegmentRange> segmentRangeIterator =
+                        batchClient.getSegments(Stream.of(scopeName, stream), null, null).getIterator();
+                while (segmentRangeIterator.hasNext()) {
+                    SegmentRange segmentRange = segmentRangeIterator.next();
+                    Segment segment = Segment.fromScopedName(segmentRange.getScopeName() + "/"
+                            + segmentRange.getStreamName() + "/" + segmentRange.getSegmentNumber());
```
EronWright: Seems there's a couple of Pravega issues to be reported here:
1. `SegmentRangeImpl` is not serializable and thus cannot be stored in the `PravegaInputSplit`.
2. `SegmentRange.getSegment()` returns a corrupted `Segment` object (the scope is wrong).  This forces us to reconstruct the segment object.

vijikarthi: For (1), I have created a JIRA ticket https://github.com/pravega/pravega/issues/2433

For (2), I don't see an API to get the segment reference from segment range directly `SegmentRange.getSegment()`. 

I was trying to create a new segment instance (https://github.com/pravega/pravega/blob/master/client/src/main/java/io/pravega/client/segment/impl/Segment.java#L38) which apparently was not working (corrupted segment). The issue got resolved when I try to create the segment using `Segment.fromScopedName()` API.

fpj: (1) I think this PR solves the issue, yes? https://github.com/pravega/pravega/pull/2434

(2) I don't see any specific reason for requiring  a `Segment` object in `PravegaInputStream`. Perhaps we should simply replace segment `Segment` with  `SegmentRange` in the constructor and internally. Would it work?

vijikarthi: On (2), yes I was using Segment instance because of (1). Now that (1) is resolved, I will pull the change and modify `PravegaInputStream` to use only `SegmentRange`

EronWright:
@vijikarthi I updated the dependencies based on some testing with SSL on Windows.   Also made the corresponding updates to `pravega-samples` in `develop` branch.

Issue #68, opened 2017-11-15
Develop a unit test for FlinkPravegaReader


EronWright:
**Problem description**
The `FlinkPravegaReader` has a test `FlinkPravegaReaderTest` that is best understood as an integration test.   Coverage is too low.

**Suggestions for an improvement**
- Rename `FlinkPravegaReader` to `FlinkPravegaReaderITCase`.
- Develop a unit test that uses a mock Pravega `EventStreamReader`.

tzulitai:
@EronWright, I would gladly take this issue.

Also, I assume that this together with #67 should subsume #51?

EronWright:
@tzulitai sounds great, would you mind following the style seen in #76 ?

tzulitai:
Sure, no problem 👌

EronWright:
This is fixed

Issue #62, opened 2017-10-12
Implement table source (batch)


EronWright:
**Problem description**
As a follow-up to #32, extend `FlinkPravegaTableSource` to support the batch API, based on the underlying `FlinkPravegaInputFormat`. 

Please note that Flink's in-built `CsvTableSource` implements **both** `StreamTableSource` and `BatchTableSource`.  We want the same here.

EronWright:
This was fixed in #125.

Issue #124, opened 2018-04-25
Implement table sink (batch)


EronWright:
**Problem description**
As a follow-up to #62, develop a batch table _sink_.   This allows Flink batch API to write a table to a stream.

The implementation will involve two aspects:
1. Develop a `FlinkOutputFormat` which writes a `DataSet` to a stream.    We need to understand any requirements re ordering and keys. 
2. Add the `BatchTableSink` trait to the `FlinkPravegaTableSink`, which wraps the use of (1).

vijikarthi:
closed https://github.com/pravega/flink-connectors/pull/183

PR #125, opened 2018-04-28
[issue-48] Builder API

### LINKED ISSUE #48: Use a builder for creating Pravega Reader/Writer
**Problem description**

During discussion for #30 we noticed the number of options to the constructors required for flink. It may be a nicer experience to have a builder to create readers and writers instead of using constructors. This may lead to better code reuse.

Look into using a builder for construction of the flink connectors

EronWright:
Signed-off-by: Eron Wright <eronwright@gmail.com>

**Change log description**
```
- introduced PravegaConfig to capture high-level client configuration
- general stream cut support
- support for environment-provided default Pravega scope

- builder for FlinkPravegaInputFormat

- builder for FlinkPravegaReader
- rename FlinkPravegaReaderTest to FlinkPravegaReaderITCase
- unit test for FlinkPravegaReader

- implement batch table API support
- builder for FlinkPravegaTableSource
- introduce FlinkPravegaJsonTableSource
- treat FlinkPravegaTableSource as an internal class

- deprecated StreamId in favor of io.pravega.client.stream.Stream
- deprecated FlinkPravegaParams in favor of PravegaConfig, builder classes
- moved IntegerDeserializationSchema to utils
- implemented StreamSourceOperatorTestHarness
```
**Purpose of the change**
Closes #48 (builder API), #68 (reader unit test), and #62 (TableSource batch API)

**What the code does**
This PR introduces a new builder API for the stream, batch, and table connectors, and (trivially) completes the table batch API integration.   The builder API expands the capabilities of the connector to include support for credentials and stream cuts across all connectors.  It also supports environment variables/system properties for configuring the controller URI and the default scope.

Note this is a **breaking change**.

Some usage examples:

Common:
```
PravegaConfig config = PravegaConfig.fromDefaults()
    .withControllerURI("tcp://...")
    .withCredentials(credentials)
    .withDefaultScope("scope-1");
```

Streaming API:
```
FlinkPravegaReader<Integer> pravegaSource = FlinkPravegaReader.<Integer>builder()
                    .forStream("stream-1")
                    .withPravegaConfig(config)
                    .withDeserializationSchema(...)
                    .build();
```

Batch API:
```
FlinkPravegaInputFormat inputFormat = FlinkPravegaInputFormat.<Integer>builder()
                            .forStream("stream-1")
                            .forStream(Stream.of("scope-2/stream-2"), streamCut)
                            .withPravegaConfig(config)
                            .withDeserializationSchema(...)
                            .build(),
```

Table API:
```
TableSchema schema = ...;
FlinkPravegaTableSource source = FlinkPravegaJsonTableSource.builder()
                .forStream(stream)
                .withPravegaConfig(config)
                .withSchema(schema)
                .build();
```

**How to verify it**
New and revised unit tests, notably:
- `FlinkPravegaReaderTest`
- `PravegaConfigTest`
- `FlinkPravegaTableSourceTest`

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/AbstractReaderBuilder.java
+++ b/src/main/java/io/pravega/connectors/flink/AbstractReaderBuilder.java
@@ -1,0 +1,141 @@
+/**
+ * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.connectors.flink;
+
+import io.pravega.client.stream.Stream;
+import io.pravega.client.stream.StreamCut;
+import io.pravega.connectors.flink.util.StreamWithBoundaries;
+import lombok.Data;
+import org.apache.flink.util.Preconditions;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * A base builder for connectors that consume a Pravega stream.
+ * @param <B> the builder class.
+ */
+public abstract class AbstractReaderBuilder<B extends AbstractReaderBuilder> implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private final List<StreamSpec> streams;
+
+    private PravegaConfig pravegaConfig;
+
+    protected AbstractReaderBuilder() {
+        this.streams = new ArrayList<>(1);
+        this.pravegaConfig = PravegaConfig.fromDefaults();
+    }
+
+    /**
+     * Set the Pravega client configuration, which includes connection info, security info, and a default scope.
+     *
+     * The default client configuration is obtained from {@code PravegaConfig.fromDefaults()}.
+     * @param pravegaConfig the configuration to use.
+     */
+    public B withPravegaConfig(PravegaConfig pravegaConfig) {
+        this.pravegaConfig = pravegaConfig;
+        return builder();
+    }
+
+    /**
+     * Add a stream and its associated start {@link StreamCut} to be read by the readers of a ReaderGroup.
+     * @param streamSpec the unqualified or qualified name of the stream.
+     * @param startStreamCut Start {@link StreamCut}
+     * @return A builder to configure and create a reader.
+     */
+    public B forStream(final String streamSpec, final StreamCut startStreamCut) {
+        streams.add(StreamSpec.of(streamSpec, startStreamCut, StreamCut.UNBOUNDED));
+        return builder();
+    }
+
+    /**
+     * Add a stream that needs to be read by the readers of a ReaderGroup. The current starting position of the stream
+     * will be used as the starting StreamCut.
+     * @param streamSpec the unqualified or qualified name of the stream.
+     * @return A builder to configure and create a reader.
+     */
+    public B forStream(final String streamSpec) {
+        return forStream(streamSpec, StreamCut.UNBOUNDED);
+    }
+
+    /**
+     * Add a stream and its associated start {@link StreamCut} to be read by the readers of a ReaderGroup.
+     * @param stream Stream.
+     * @param startStreamCut Start {@link StreamCut}
+     * @return A builder to configure and create a reader.
+     */
+    public B forStream(final Stream stream, final StreamCut startStreamCut) {
+        streams.add(StreamSpec.of(stream, startStreamCut, StreamCut.UNBOUNDED));
+        return builder();
+    }
+
+    /**
+     * Add a stream that needs to be read by the readers of a ReaderGroup. The current starting position of the stream
+     * will be used as the starting StreamCut.
+     * @param stream Stream.
+     * @return A builder to configure and create a reader.
+     */
+    public B forStream(final Stream stream) {
+        return forStream(stream, StreamCut.UNBOUNDED);
+    }
+
+    /**
+     * Gets the Pravega configuration.
+     */
+    protected PravegaConfig getPravegaConfig() {
+        Preconditions.checkState(pravegaConfig != null, "A Pravega configuration must be supplied.");
+        return pravegaConfig;
+    }
+
+    /**
+     * Resolves the streams to be provided to the reader, based on the configured default scope.
+     */
+    protected List<StreamWithBoundaries> resolveStreams() {
+        Preconditions.checkState(!streams.isEmpty(), "At least one stream must be supplied.");
+        PravegaConfig pravegaConfig = getPravegaConfig();
+        return streams.stream()
+                .map(s -> StreamWithBoundaries.of(pravegaConfig.resolve(s.streamSpec), s.from, s.to))
+                .collect(Collectors.toList());
+    }
+
+    protected abstract B builder();
+
+    /**
+     * A Pravega stream with optional boundaries based on stream cuts.
+     */
+    @Data
+    public static class StreamSpec implements Serializable {
+
+        private static final long serialVersionUID = 1L;
+
+        private final String streamSpec;
+        private final StreamCut from;
+        private final StreamCut to;
+
+        public static StreamSpec of(String streamSpec, StreamCut from, StreamCut to) {
+            Preconditions.checkNotNull(streamSpec, "streamSpec");
+            Preconditions.checkNotNull(streamSpec, "from");
+            Preconditions.checkNotNull(streamSpec, "to");
+            return new StreamSpec(streamSpec, from, to);
+        }
+
+        public static StreamSpec of(Stream stream, StreamCut from, StreamCut to) {
+            Preconditions.checkNotNull(stream, "stream");
+            Preconditions.checkNotNull(stream, "from");
+            Preconditions.checkNotNull(stream, "to");
+            return new StreamSpec(stream.getScopedName(), from, to);
+        }
+    }
+
+}
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -10,7 +10,7 @@
 package io.pravega.connectors.flink;
 
 import com.google.common.base.Preconditions;
-
+import io.pravega.client.ClientConfig;
 import io.pravega.client.admin.ReaderGroupManager;
 import io.pravega.client.stream.Checkpoint;
 import io.pravega.client.stream.EventRead;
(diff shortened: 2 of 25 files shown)


fpj:
There is an unused import that is causing checkstyle to fail:

```
:checkstyleMain[ant:checkstyle] [ERROR] /home/travis/build/pravega/flink-connectors/src/main/java/io/pravega/connectors/flink/CheckpointSerializer.java:13:8: Unused import - org.apache.commons.lang3.SerializationUtils. [UnusedImports]
```

EronWright:
The findbugs plugin is complaining due to some recent changes in Pravega.  Am fixing.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -320,23 +223,240 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+        FlinkPravegaReader<T> buildSourceFunction() {
+
+            // rgConfig
+            ReaderGroupConfig.ReaderGroupConfigBuilder rgConfigBuilder = ReaderGroupConfig
+                    .builder()
+                    .disableAutomaticCheckpoints();
+            if (this.readerGroupRefreshTime != null) {
+                rgConfigBuilder.groupRefreshTimeMillis(this.readerGroupRefreshTime.toMilliseconds());
+            }
+            resolveStreams().forEach(s -> rgConfigBuilder.stream(s.getStream(), s.getFrom() /*, s.getTo() */));
```
vijikarthi: Why is the `to` stream cut commented and left to default UNBOUNDED?

src/test/java/io/pravega/connectors/flink/FlinkPravegaJsonTableSourceTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import io.pravega.connectors.flink.serialization.JsonRowDeserializationSchema;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeinfo.Types;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.table.api.TableSchema;
+import org.apache.flink.types.Row;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
vijikarthi: Please replace the bulk import statement with specific package imports.

vijikarthi:
Since this patch introduces breaking API changes, we need to fix the connector dependent projects (nautilus-samples, pravega-samples etc.,) before merging this code.

EronWright:
@vijikarthi OK this is ready for another review.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -129,95 +96,37 @@ public FlinkPravegaReader(final URI controllerURI, final String scope, final Set
...
-     * <p>This timeout is the value passed to {@link EventStreamReader#readNextEvent(long)}.
-     * 
-     * @return The timeout, in milliseconds
+     * Initializes the reader.
      */
-    public long getEventReadTimeout() {
-        return eventReadTimeout;
+    void initialize() {
+        // TODO: This will require the client to have access to the pravega controller and handle any temporary errors.
+        //       See https://github.com/pravega/pravega/issues/553.
```
fpj: This issue is actually closed. If this is still a concern, then we need an issue against this repository.

I'm wondering if this is an actual concern, though. On the one hand, if the controller is not accessible when this environment is being constructed, then it is a good idea to fail early. On the other hand, this is just setting up configuration for submission, and creating a reader group is really part of runtime. 

It also depends on when this `initialize` call is actually invoked. If at job submission or job execution time, then we are fine and we don't need another issue.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -320,23 +223,244 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+
+    /**
+     * An abstract streaming reader builder.
+     *
+     * The builder is abstracted to act as the base for both the {@link FlinkPravegaReader} and {@link FlinkPravegaTableSource} builders.
+     *
+     * @param <T> the element type.
+     * @param <B> the builder type.
+     */
+    static abstract class AbstractStreamingReaderBuilder<T, B extends AbstractStreamingReaderBuilder> extends AbstractReaderBuilder<B> {
```
fpj: Is this class better in a separate file for organization purposes?

src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
```diff
@@ -78,42 +70,84 @@ public FlinkPravegaTableSource(
...
     }
 
+    /**
+     * NOTE: This method is for internal use only for defining a TableSource.
+     *       Do not use it in Table API programs.
+     */
     @Override
-    public TypeInformation<Row> getReturnType() {
-        return typeInfo;
+    public DataSet<Row> getDataSet(ExecutionEnvironment env) {
```
fpj: If this method is coming from the implemented interfaces, then why is it the case that programs can't use it? I'd say that this needs to be at least package protected, but the compiler won't allow you to do it.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -764,4 +734,130 @@ public String getStream() {
...
+
+    /**
+     * An abstract streaming writer builder.
+     *
+     * The builder is abstracted to act as the base for both the {@link FlinkPravegaWriter} and {@link FlinkPravegaTableSink} builders.
+     *
+     * @param <T> the element type.
+     * @param <B> the builder type.
+     */
+    public abstract static class AbstractStreamingWriterBuilder<T, B extends AbstractStreamingWriterBuilder> extends AbstractWriterBuilder<B> {
```
fpj: Same question about being in a separate file.

src/main/java/io/pravega/connectors/flink/util/StreamId.java
```diff
@@ -1,77 +0,0 @@
...
-
-import org.apache.commons.lang3.StringUtils;
-
-import java.util.Objects;
-
-/**
- * Captures the fully qualified name of a stream. The convention to represent this as a
- * single string is using [scope]/[stream].
- */
-public class StreamId {
```
fpj: The description says that this is being deprecated, but the class is actually being removed rather than marked as deprecated.

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderTest.java
```diff
@@ -9,186 +9,331 @@
...
-
-    // Number of events to produce into the test stream.
-    private static final int NUM_STREAM_ELEMENTS = 10000;
-
-    // Setup utility.
-    private static final SetupUtils SETUP_UTILS = new SetupUtils();
-
-    //Ensure each test completes within 120 seconds.
-    @Rule
-    public final Timeout globalTimeout = new Timeout(120, TimeUnit.SECONDS);
```
fpj: It looks like these test cases have been left timeout-less.

src/test/java/io/pravega/connectors/flink/FlinkTableITCase.java
```diff
@@ -146,6 +155,63 @@ public void testEndToEnd() throws Exception {
         }
     }
 
+
+    /**
+     * Tests the end-to-end functionality of a batch table source & sink.
```
fpj: & -> and

src/test/java/io/pravega/connectors/flink/FlinkTableITCase.java
```diff
@@ -146,6 +155,63 @@ public void testEndToEnd() throws Exception {
...
+     * The test then uses the {@link FlinkPravegaTableSource} to absorb the changelog as a new table.
+     *
+     * <p>Flink's ability to convert POJOs (e.g. {@link SampleRecord}) to/from table rows is also demonstrated.
+     *
+     * <p>Because the source is unbounded, the test must throw an exception to deliberately terminate the job.
+     *
+     * @throws Exception on exception
+     */
+    @Test
+    @Ignore("[issue-124] FlinkPravegaTableSink doesn't support BatchTableSink")
```
fpj: Instead of ignoring the test case, should we remove and add it back once 124 is done? It can go in the same PR.

src/main/java/io/pravega/connectors/flink/AbstractReaderBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+    @Data
+    private static class StreamSpec implements Serializable {
+
+        private static final long serialVersionUID = 1L;
+
+        private final String streamSpec;
+        private final StreamCut from;
+        private final StreamCut to;
+
+        public static StreamSpec of(String streamSpec, StreamCut from, StreamCut to) {
```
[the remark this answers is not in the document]
EronWright: Well, the main idea here is that the 'streamSpec' is an unprocessed user input that will be parsed later (out of necessity).  I avoided 'streamName' because I don't want anyone to treat the value as an actual stream name.  Also it is a private class.

src/main/java/io/pravega/connectors/flink/CheckpointSerializer.java
```diff
@@ -32,15 +32,17 @@ public int getVersion() {
 
     @Override
     public byte[] serialize(Checkpoint checkpoint) throws IOException {
-        return SerializationUtils.serialize(checkpoint);
+        ByteBuffer buf = checkpoint.toBytes();
+        byte[] b = new byte[buf.remaining()];
```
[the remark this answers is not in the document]
EronWright: This is the way to convert an opaque `ByteBuffer` to a `byte[]`.  I would argue that `toBytes` should return a `byte[]`.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -55,69 +47,44 @@
...
     private static final long serialVersionUID = 1L;
 
-    private static final long DEFAULT_EVENT_READ_TIMEOUT = 1000;
-
-    private static final long DEFAULT_CHECKPOINT_INITIATE_TIMEOUT = 5000;
-
     // ----- configuration fields -----
 
-    // The supplied event deserializer.
-    private final DeserializationSchema<T> deserializationSchema;
```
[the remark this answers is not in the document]
EronWright: For the sake of the new unit test that makes assertions against them.

src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
EronWright: I agree that it's a questionable design choice on the part of Flink Table API, but there's nothing I can do.

src/test/java/io/pravega/connectors/flink/FlinkTableITCase.java
EronWright: I would prefer we leave it, since these integration tests are illustrative and we don't have any other samples.   It will be fixed soon in #124

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -320,23 +223,244 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+     * Create the {@link ReaderGroup} for the current configuration.
      */
-    private ReaderGroup createReaderGroup() {
-        Map<Stream, StreamCut> streamConfigMap = new HashMap<>();
-        for (String stream: streamNames) {
-            streamConfigMap.put(Stream.of(scopeName, stream), StreamCut.UNBOUNDED);
+    protected ReaderGroup createReaderGroup() {
+        ReaderGroupManager readerGroupManager = createReaderGroupManager();
+        readerGroupManager.createReaderGroup(this.readerGroupName, readerGroupConfig);
+        return readerGroupManager.getReaderGroup(this.readerGroupName);
```
shrids: We should invoke readerGroupManager.close() (since closing the `connectionFactory` does not close the connection to controller)

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
fpj: That's fine, but let's at least replace the issue from the pravega queue, which is currently closed, with one in the flink-connector queue. I have created this one:

https://github.com/pravega/flink-connectors/issues/130

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
fpj: Yeah, good point. The reader group object needs to be closed after this merge:

https://github.com/pravega/pravega/pull/2383/files#diff-67be2af2da7593612509f695490a147bR150

Looking at the checkpoint hook, it is not clear to me where it needs to happen, though.

EronWright: I too have been worried about the cleanup issue, so I added a `close` method to Flink's hook interface in 1.5.

https://github.com/apache/flink/blob/af29172bf555558e579a92f0ee3e56ed76fd7a1a/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/MasterTriggerRestoreHook.java#L80

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderTest.java
EronWright: The unit tests don't wait for anything in an unbounded way that would necessitate a timeout.   I'll review whether any of integration tests (which I agree should use a timeout) are lacking a timeout.

EronWright:
@fpj and @shrids I think I've addressed all feedback.

PR #132, opened 2018-05-14
[issue-122] run tests with secure Pravega cluster (WIP)

### LINKED ISSUE #122: Add test case(s) to run and test connector with secure Pravega cluster
**Problem description**

* Add test cases that runs `InProcPravegaCluster` in `secure` mode and validate connectors implementation for read/write/stream management path. 

**Problem location**
Flink connector system/integration test

hldnova:
**Change log description**
- add reader test against InProcPravegaCluster with authentication enabled.
- add writer test against InProcPravegaCluster with authentication enabled.

Note that TLS tests not added because a bug in Pravega. Will create a separate issue to track TLS tests.

**Purpose of the change**

**What the code does**
This PR added reader/writer tests against Pravega with authentication enabled.

**How to verify it**
Run the flink connector tests, esp. FlinkPravegaReaderEnableAuthITCase, and FlinkPravegaWriterEnableAuthITCase.


src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderITCase.java
```diff
@@ -40,21 +40,22 @@
...
     public final Timeout globalTimeout = new Timeout(120, TimeUnit.SECONDS);
-    
-    @BeforeClass
-    public static void setupPravega() throws Exception {
-        SETUP_UTILS.startAllServices();
+
+    // Setup utility.
+    protected SetupUtils setupUtils = new SetupUtils();
+
+    @Before
```
vijikarthi: Is there any reason why we we change this from @BeforeClass to @Before?

src/test/java/io/pravega/connectors/flink/utils/SetupUtils.java
```diff
@@ -234,7 +262,7 @@ public void createTestStream(final String streamName, final int numSegments)
         ClientConfig getClientConfig();
     }
 
-    static class InProcPravegaGateway implements PravegaGateway {
+    class InProcPravegaGateway implements PravegaGateway {
```
vijikarthi: Can we have a separate Gateway for secure cluster `SecureInProcPravegaGateway extends InProcPravegaGateway` and use this for running the secure tests?

hldnova: There seems no need for that as InProcPravegaCluster now requires builder to provide the secure related parameters, user/password, key file, etc. Otherwise, it would throw NPE.

I will take a look at this again to see if it make senses in the context of reverting SetupUtils back to static.

hldnova:
a new one will be submitted

PR #133, opened 2018-05-15
[issue-122] run tests with secure Pravega cluster

### LINKED ISSUE #122: Add test case(s) to run and test connector with secure Pravega cluster

hldnova:
**Change log description**
- add reader test against InProcPravegaCluster with authentication enabled.
- add writer test against InProcPravegaCluster with authentication enabled.

Note that TLS tests not added because a bug in Pravega. Will create a separate issue to track TLS tests.

**Purpose of the change**
closes #122 

**What the code does**
This PR added reader/writer tests against Pravega with authentication enabled.

**How to verify it**
Run the flink connector tests, esp. FlinkPravegaReaderEnableAuthITCase, and FlinkPravegaWriterEnableAuthITCase.


src/test/java/io/pravega/connectors/flink/utils/SetupUtils.java
```diff
@@ -304,4 +328,22 @@ public ClientConfig getClientConfig() {
                     .build();
         }
     }
+
+    /*
+     * A utility class to set flags for tests against secure Pravega.
+     */
+    public static class SetupUtilsUpdate {
```
EronWright: Why not simply enhance `SetupUtils` to provide this functionality directly?  Like, provide public setters that must be called before `.start()`.

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderEnableAuthITCase.java
```diff
@@ -0,0 +1,26 @@
...
+package io.pravega.connectors.flink;
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.BeforeClass;
+
+/**
+ * Integration tests for {@link FlinkPravegaWriter}.
+ */
+@Slf4j
+public class FlinkPravegaReaderEnableAuthITCase extends FlinkPravegaReaderITCase {
```
EronWright: I don't understand the purpose of this test.

Note: the javadoc comments above are wrong.

src/test/java/io/pravega/connectors/flink/utils/SetupUtils.java
```diff
@@ -47,12 +49,20 @@
 public final class SetupUtils {
 
     private static final ScheduledExecutorService DEFAULT_SCHEDULED_EXECUTOR_SERVICE = ExecutorServiceHelpers.newScheduledThreadPool(3, "SetupUtils");
+    private static final String PRAVEGA_USERNAME = "admin";
```
EronWright: These file paths make me nervous because they presuppose the working directory of the test.   

Ideally we would not depend on the pravega submodule for the below files.  The best solution would be to place them into `src/test/resources` and extract them as temporary files.   

Could do something like:
```
import java.nio.file.Files;
public static Path copyResourceToFile(String resourceName) throws IOException {
    Path tempPath = Files.createTempFile("test-", ".tmp");
    tempPath.toFile().deleteOnExit();
    try (InputStream stream = ResourceManager.class.getResourceAsStream(jarFilePath)) {
        Files.copy(ResourceManager.class.getResourceAsStream(resourceName), tempPath, StandardCopyOption.REPLACE_EXISTING);
    }
    return tempPath;
}
```

src/test/java/io/pravega/connectors/flink/utils/SetupUtils.java
hldnova: Done.

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderEnableAuthITCase.java
hldnova: It is supposed to re-run the tests in FlinkPravegaReaderITCase, but with Pravega running in secure mode. Same for the writer test.

src/test/java/io/pravega/connectors/flink/utils/SetupUtils.java
hldnova: Done. Thanks for the tip! Somehow had trouble creating stream using ResourceManager. Used getClass().getClassLoader() instead.

EronWright:
I would like to avoid the running the secure and insecure tests until it is warranted.  My suggestion is:

1. drop the "EnableAuth" variants 
2. in the `setupPravega` methods of the existing ITCase classes, do this:
```
SETUP_UTILS.setEnableAuth(true);
SETUP_UTILS.startAllServices();
```
(so that we don't need the awkward `startSecureServices`)
or, add an overloaded `SetupUtils(bool enableTLS, bool enableAuth)` constructor (edited)
3. be sure to call `.credentials(new DefaultCredentials(PRAVEGA_PASSWORD, PRAVEGA_USERNAME))` in the `ExternalPravegaGateway` variant (edited)

hldnova:
@EronWright, your suggestions had been implemented.

Issue #126, opened 2018-05-02
Update default transaction timeouts in FlinkPravegaWriter to sync with Pravega defaults


vijikarthi:
**Problem description**
For the Exactly once writer, the default transaction grace period timeout is more than the allowed limit.

```
java.util.concurrent.CompletionException: io.pravega.shaded.io.grpc.StatusRuntimeException: INTERNAL: scaleGracePeriod too large, max value is 30000
```

**Problem location**
`FlinkPravegaWriter` implementation

https://github.com/pravega/flink-connectors/blob/master/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java#L64

fpj:
This timeout will actually go away once we implement rolling txns.

vijikarthi:
closed https://github.com/pravega/flink-connectors/pull/147

PR #147, opened 2018-05-29
[issue-126] Update default transaction timeouts in FlinkPravegaWriter to sync with Pravega defaults

vijikarthi:

**Change log description**
This PR addresses the issue https://github.com/pravega/flink-connectors/issues/126
- The transaction timeouts are updated to use the maximum defaults of Pravega

**Purpose of the change**
- To avoid flink app from failing due to larger transaction timeouts

**What the code does**
- configures default timeouts when the value is not supplied

**How to verify it**

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/AbstractStreamingWriterBuilder.java
+++ b/src/main/java/io/pravega/connectors/flink/AbstractStreamingWriterBuilder.java
@@ -24,14 +24,14 @@ import org.apache.flink.util.Preconditions;
 public abstract class AbstractStreamingWriterBuilder<T, B extends AbstractStreamingWriterBuilder> extends AbstractWriterBuilder<B> {
 
     // the numbers below are picked based on the default max settings in Pravega
-    private static final long DEFAULT_TXN_TIMEOUT_MILLIS = 30000; // 30 seconds
+    private static final long DEFAULT_TXN_LEASE_RENEWAL_PERIOD_MILLIS = 30000; // 30 seconds
 
     protected PravegaWriterMode writerMode;
-    protected Time txnTimeout;
+    protected Time txnLeaseRenewalPeriod;
 
     protected AbstractStreamingWriterBuilder() {
         writerMode = PravegaWriterMode.ATLEAST_ONCE;
-        txnTimeout = Time.milliseconds(DEFAULT_TXN_TIMEOUT_MILLIS);
+        txnLeaseRenewalPeriod = Time.milliseconds(DEFAULT_TXN_LEASE_RENEWAL_PERIOD_MILLIS);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/io/pravega/connectors/flink/AbstractStreamingWriterBuilder.java
```diff
@@ -24,8 +24,8 @@
 public abstract class AbstractStreamingWriterBuilder<T, B extends AbstractStreamingWriterBuilder> extends AbstractWriterBuilder<B> {
 
     // the numbers below are picked somewhat arbitrarily at this point
```
EronWright: Is the above comment in the code still true?   Are these values consistent with the underlying Pravega client's defaults?  The new values are radically different, which suggests a problem, i.e. that we don't really understand them.

vijikarthi: > // the numbers below are picked somewhat arbitrarily at this point

I forgot to remove this comment (will update). 

Yes, the values are consistent with Pravega's default and  I have posted my observation here https://github.com/pravega/flink-connectors/issues/126

src/main/java/io/pravega/connectors/flink/AbstractStreamingWriterBuilder.java
```diff
@@ -23,9 +23,9 @@
  */
 public abstract class AbstractStreamingWriterBuilder<T, B extends AbstractStreamingWriterBuilder> extends AbstractWriterBuilder<B> {
 
-    // the numbers below are picked somewhat arbitrarily at this point
-    private static final long DEFAULT_TXN_TIMEOUT_MILLIS = 2 * 60 * 60 * 1000; // 2 hours
-    private static final long DEFAULT_TX_SCALE_GRACE_MILLIS = 10 * 60 * 1000; // 10 minutes
+    // the numbers below are picked based on the default max settings in Pravega
+    private static final long DEFAULT_TXN_TIMEOUT_MILLIS = 30000; // 30 seconds
```
fpj: This timeout value corresponds to the lease period of a transaction. The writer renews the lease within the configured period, and it does it automatically, without any necessary intervention from the application. 30s sounds like a pretty reasonable default value to me.

This configuration parameter also determines the maximum amount of time we allow a txn to remain open. The maximum execution time is calculated as the minimum of 1 day and 1000x the lease period.

src/main/java/io/pravega/connectors/flink/AbstractStreamingWriterBuilder.java
```diff
@@ -23,9 +23,9 @@
...
+    private static final long DEFAULT_TX_SCALE_GRACE_MILLIS = 30000; // 30 seconds
```
fpj: The scale grace period is a parameter that is supposed to go away with the rolling txns feature. Currently, it cannot be smaller than the lease period.

EronWright:
Here's the two options that I favor:

1. Rename `withTxnTimeout` to `withTxnLeaseRenewalPeriod`, default value of 30 seconds:
```
    /**
     * Sets the transaction lease renewal period.
     *
     * When the writer mode is set to {@code EXACTLY_ONCE}, transactions are used to persist events to the Pravega stream.  The transaction interval corresponds to the Flink checkpoint interval.   Throughout that interval, the transaction is kept alive with a lease that is periodically renewed.    This configuration setting sets the lease renewal period.  The default value is 30 seconds.
     *
     * @param period the lease renewal period
     */
    public B withTxnLeaseRenewalPeriod(Time period)
```

2. Keep the API as-is, change the default value to 8 hours, and divide the value by 1000 when configuring the underlying Pravega client (since it works in terms of lease renewal period).

vijikarthi:
My suggestion is to merge this PR and create a separate issue post EAP since any change to the API will impact other dependencies.

vijikarthi:
Can this be merged?

PR #156, opened 2018-06-15
[issue-66] Use single threaded executor in FlinkPravegaWriter

### LINKED ISSUE #66: Use single threaded executor in FlinkPravegaWriter
**Problem description**
The non-transactional writer uses a larger thread pool than necessary for handling write completion callbacks.   

Use a single threaded executor to reduce the number of created threads and to reduce contention.  The body of the callback synchronizes on the internal writer anyway, so having numerous threads simply increases contention.

EronWright:
Signed-off-by: Eron Wright <eronwright@gmail.com>

**Change log description**
- Use single threaded executor in FlinkPravegaWriter

**Purpose of the change**
Avoid wasting threads when a single thread will suffice.  Closes #66.

**What the code does**
Reduces the non-transactional writer's thread pool size from 5 to 1.

**How to verify it**
- FlinkPravegaWriter

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
@@ -223,7 +223,7 @@ public class FlinkPravegaWriter<T>
 
     @VisibleForTesting
     protected ExecutorService createExecutorService() {
-        return Executors.newFixedThreadPool(5);
+        return Executors.newSingleThreadExecutor();
     }
 
     private void initializeInternalWriter() {


src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -223,7 +223,7 @@ protected AbstractInternalWriter createInternalWriter() {
 
     @VisibleForTesting
     protected ExecutorService createExecutorService() {
-        return Executors.newFixedThreadPool(5);
+        return Executors.newSingleThreadExecutor();
```
fpj: One clarifying question. This executor is used with the non-transactional writer with a `whenCompleteAsync`. Can we restructure it to use the default thread pool instead? I'm just wondering if we really need this executor in the first place. Otherwise, the change looks good.

EronWright: I think we would best pull in whomever wrote the original code to get their perspective on that.    And I would prefer we do that later.  For now, I propose we go with this simple improvement.

fpj: Ok, I have created an issue for it: #157

PR #169, opened 2018-07-19
[issue-31] Implement connector metrics

### LINKED ISSUE #31: Implement connector metrics
**Problem description**
The connector should emit relevant metrics using Flink's metrics API.   Note that Flink already provides some basic metrics like records in/out, which needn't be duplicated.   

Suggested metrics:
- configuration data (stream names as gauges?)
- segment assignment
- event I/O rates
- position information
- lag (`ReaderGroupMetrics::unreadBytes`)

vijikarthi:

  * Added support to draw reader group metrics for FlinkPravegaReader
  * Following metrics are enabled
    - reader group name
    - scope
    - stream names
    - online readers
    - segment positions
    - unread bytes

**Change log description**
Fixes #31

**Purpose of the change**
To publish reader group metrics

**What the code does**
Publishes the runtime stats of the reader group as Flink metrics

**How to verify it**
- On a running Flink cluster alongside with Pravega, run a job that uses `FlinkPravegaReader`  (e.g., anomaly-samples application) and verify if the metrics are visible from UI as well as REST API
```
curl -i -s -f localhost:1028/jobs/2c248d6fe8ed518e959989889a3c5ced/vertices/bc764cd8ddf7a0cff126f51c16239658/metrics?get=0.Source__AnomalyEventReader.PravegaReader.readerGroup.readerGroupName
[{"id":"0.Source__AnomalyEventReader.PravegaReader.readerGroup.readerGroupName","value":"flinkmrdf01aislmcira301jg"}]

curl -i -s -f localhost:1028/jobs/2c248d6fe8ed518e959989889a3c5ced/vertices/bc764cd8ddf7a0cff126f51c16239658/metrics?get=0.Source__AnomalyEventReader.PravegaReader.readerGroup.scope
[{"id":"0.Source__AnomalyEventReader.PravegaReader.readerGroup.scope","value":"anomaly"}]

curl -i -s -f localhost:1028/jobs/2c248d6fe8ed518e959989889a3c5ced/vertices/bc764cd8ddf7a0cff126f51c16239658/metrics?get=0.Source__AnomalyEventReader.PravegaReader.readerGroup.streams
[{"id":"0.Source__AnomalyEventReader.PravegaReader.readerGroup.streams","value":"test"}]

curl -i -s -f localhost:1028/jobs/2c248d6fe8ed518e959989889a3c5ced/vertices/bc764cd8ddf7a0cff126f51c16239658/metrics?get=0.Source__AnomalyEventReader.PravegaReader.readerGroup.onlineReaders
[{"id":"0.Source__AnomalyEventReader.PravegaReader.readerGroup.onlineReaders","value":"Source: AnomalyEventReader (1/1)"}]

curl -i -s -f localhost:1028/jobs/2c248d6fe8ed518e959989889a3c5ced/vertices/bc764cd8ddf7a0cff126f51c16239658/metrics?get=0.Source__AnomalyEventReader.PravegaReader.readerGroup.stream.test.segmentPositions
[{"id":"0.Source__AnomalyEventReader.PravegaReader.readerGroup.stream-test.segmentPositions","value":"stream=test, segments={anomaly/test/2.#epoch.0=0 ,anomaly/test/0.#epoch.0=0 ,anomaly/test/1.#epoch.0=0 ,}"}]

curl -i -s -f localhost:1028/jobs/2c248d6fe8ed518e959989889a3c5ced/vertices/bc764cd8ddf7a0cff126f51c16239658/metrics?get=0.Source__AnomalyEventReader.PravegaReader.readerGroup.unreadBytes
[{"id":"0.Source__AnomalyEventReader.PravegaReader.readerGroup.unreadBytes","value":"3197"}]

```

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/AbstractReaderBuilder.java
+++ b/src/main/java/io/pravega/connectors/flink/AbstractReaderBuilder.java
@@ -33,6 +33,8 @@ public abstract class AbstractReaderBuilder<B extends AbstractReaderBuilder> imp
 
     private PravegaConfig pravegaConfig;
 
+    private boolean enableMetrics = true;
+
     protected AbstractReaderBuilder() {
         this.streams = new ArrayList<>(1);
         this.pravegaConfig = PravegaConfig.fromDefaults();
@@ -135,6 +137,24 @@ public abstract class AbstractReaderBuilder<B extends AbstractReaderBuilder> imp
         return streams.stream()
                 .map(s -> StreamWithBoundaries.of(pravegaConfig.resolve(s.streamSpec), s.from, s.to))
                 .collect(Collectors.toList());
+    }
+
+    /**
+     * enable/disable pravega reader metrics (default: enabled).
+     *
+     * @param enable boolean
+     * @return A builder to configure and create a reader.
+     */
+    public B enableMetrics(boolean enable) {
+        this.enableMetrics = enable;
+        return builder();
+    }
+
+    /**
+     * getter to fetch the metrics flag.
+     */
+    protected boolean isMetricsEnabled() {
+        return enableMetrics;
     }
 
     protected abstract B builder();
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -37,6 +37,7 @@ import org.apache.flink.util.FlinkException;
 
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import static io.pravega.connectors.flink.util.FlinkPravegaUtils.createPravegaReader;
@@ -360,8 +361,7 @@ public class FlinkPravegaReader<T>
 
         @Override
         public String getValue() {
-            return readerGroup.getStreamCuts().keySet().stream()
-                    .map(Stream::getScopedName).collect(Collectors.joining(","));
+            return readerGroup.getStreamNames().stream().collect(Collectors.joining(","));
         }
     }
 
@@ -413,15 +413,15 @@ public class FlinkPravegaReader<T>
         readerGroupMetricGroup.gauge(ONLINE_READERS_METRICS_GAUGE, new OnlineReadersGauge(readerGroup));
         readerGroupMetricGroup.gauge(STREAM_NAMES_METRICS_GAUGE, new StreamNamesGauge(readerGroup));
 
-        Map<Stream, StreamCut> streamCuts = readerGroup.getStreamCuts();
-        for (Map.Entry<Stream, StreamCut> entry: streamCuts.entrySet()) {
-            Stream stream = entry.getKey();
+        Set<String> streamNames = readerGroup.getStreamNames();
+        for (String scopedStream: streamNames) {
+            String[] streamInfo = scopedStream.split("/", 2);
+            Preconditions.checkArgument(streamInfo.length == 2, "not a fully qualified stream expected: scopeName/streamName");
             MetricGroup streamMetricGroup = readerGroupMetricGroup
-                    .addGroup(STREAM_METRICS_GROUP + "." + stream.getScope()+ "_"+ stream.getStreamName());
+                    .addGroup(STREAM_METRICS_GROUP + "." + streamInfo[0]+ "_"+ streamInfo[1]);
             streamMetricGroup.gauge(SEGMENT_POSITIONS_METRICS_GAUGE,
-                    new SegmentPositionsGauge(readerGroup, stream.getScope(), stream.getStreamName()));
+                    new SegmentPositionsGauge(readerGroup, streamInfo[0], streamInfo[1]));
         }
-
     }
 
     // ------------------------------------------------------------------------
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -18,17 +18,25 @@ import io.pravega.client.stream.EventStreamReader;
 import io.pravega.client.stream.ReaderConfig;
 import io.pravega.client.stream.ReaderGroup;
 import io.pravega.client.stream.ReaderGroupConfig;
+import io.pravega.client.stream.Stream;
+import io.pravega.client.stream.StreamCut;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.flink.api.common.functions.StoppableFunction;
 import org.apache.flink.api.common.serialization.DeserializationSchema;
 import org.apache.flink.api.common.time.Time;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.java.typeutils.ResultTypeQueryable;
+import org.apache.flink.metrics.Gauge;
+import org.apache.flink.metrics.MetricGroup;
 import org.apache.flink.runtime.checkpoint.MasterTriggerRestoreHook;
 import org.apache.flink.streaming.api.checkpoint.ExternallyInducedSource;
 import org.apache.flink.streaming.api.datastream.SingleOutputStreamOperator;
 import org.apache.flink.streaming.api.functions.source.RichParallelSourceFunction;
 import org.apache.flink.util.FlinkException;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 import static io.pravega.connectors.flink.util.FlinkPravegaUtils.createPravegaReader;
 
@@ -43,6 +51,26 @@ public class FlinkPravegaReader<T>
         implements ResultTypeQueryable<T>, StoppableFunction, ExternallyInducedSource<T, Checkpoint> {
 
     private static final long serialVersionUID = 1L;
+
+    // ----- metrics field constants -----
+
+    private static final String PRAVEGA_READER_METRICS_GROUP = "PravegaReader";
+
+    private static final String READER_GROUP_METRICS_GROUP = "readerGroup";
+
+    private static final String STREAM_METRICS_GROUP = "stream";
+
+    private static final String UNREAD_BYTES_METRICS_GAUGE = "unreadBytes";
+
+    private static final String READER_GROUP_NAME_METRICS_GAUGE = "readerGroupName";
+
+    private static final String SCOPE_NAME_METRICS_GAUGE = "scope";
+
+    private static final String ONLINE_READERS_METRICS_GAUGE = "onlineReaders";
+
+    private static final String STREAM_NAMES_METRICS_GAUGE = "streams";
+
+    private static final String SEGMENT_POSITIONS_METRICS_GAUGE = "segmentPositions";
 
     // ----- configuration fields -----
 
@@ -69,6 +97,9 @@ public class FlinkPravegaReader<T>
 
     // the timeout for call that initiates the Pravega checkpoint 
     final Time checkpointInitiateTimeout;
+
+    // flag to enable/disable metrics
+    final boolean enableMetrics;
 
     // ----- runtime fields -----
 
@@ -101,10 +132,12 @@ public class FlinkPravegaReader<T>
      * @param deserializationSchema     The implementation to deserialize events from Pravega streams.
      * @param eventReadTimeout          The event read timeout.
      * @param checkpointInitiateTimeout The checkpoint initiation timeout.
+     * @param enableMetrics             Flag to indicate whether metrics needs to be enabled or not.
      */
     protected FlinkPravegaReader(String hookUid, ClientConfig clientConfig,
                                  ReaderGroupConfig readerGroupConfig, String readerGroupScope, String readerGroupName,
-                                 DeserializationSchema<T> deserializationSchema, Time eventReadTimeout, Time checkpointInitiateTimeout) {
+                                 DeserializationSchema<T> deserializationSchema, Time eventReadTimeout, Time checkpointInitiateTimeout,
+                                 boolean enableMetrics) {
 
         this.hookUid = Preconditions.checkNotNull(hookUid, "hookUid");
         this.clientConfig = Preconditions.checkNotNull(clientConfig, "clientConfig");
@@ -114,6 +147,7 @@ public class FlinkPravegaReader<T>
         this.deserializationSchema = Preconditions.checkNotNull(deserializationSchema, "deserializationSchema");
         this.eventReadTimeout = Preconditions.checkNotNull(eventReadTimeout, "eventReadTimeout");
         this.checkpointInitiateTimeout = Preconditions.checkNotNull(checkpointInitiateTimeout, "checkpointInitiateTimeout");
+        this.enableMetrics = enableMetrics;
     }
 
     /**
@@ -132,6 +166,10 @@ public class FlinkPravegaReader<T>
 
     @Override
     public void run(SourceContext<T> ctx) throws Exception {
+
+        if (enableMetrics) {
+            registerMetrics();
+        }
 
         final String readerId = getRuntimeContext().getTaskNameWithSubtasks();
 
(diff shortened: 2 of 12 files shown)


src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +258,150 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+
+        @Override
+        public String getValue() {
+            StringBuilder builder = new StringBuilder();
+            builder.append("stream=").append(stream).append(", segments={");
+            Map<Stream, StreamCut> streamCuts = readerGroup.getStreamCuts();
+            Optional<Map.Entry<Stream, StreamCut>> optionalStreamCutEntry =
+                    streamCuts.entrySet().stream().filter(e -> e.getKey().getStreamName().equals(stream)).findFirst();
+            if (optionalStreamCutEntry.isPresent()) {
+                optionalStreamCutEntry.get().getValue().asImpl().getPositions().entrySet().stream().forEach(entry -> {
```
fpj: I think it is pretty clear at this point that we need one or more string representations of a stream cut. In this particular case, we should be using a `toString` call to get it rather than doing this hack. @shrids we need to enable such call on the stream cut API.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +258,150 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
         checkpointTrigger.triggerCheckpoint(checkpointId);
     }
 
+    // ------------------------------------------------------------------------
+    //  metrics
```
fpj: I'm not sure if you want to introduce this here, but we could also add `unreadBytes` from the metrics interface. We should be adding more metrics in the future:

https://github.com/pravega/pravega/blob/master/client/src/main/java/io/pravega/client/stream/ReaderGroupMetrics.java

I'm also wondering if we should have metrics for number of segments out of the listener:

https://github.com/pravega/pravega/tree/master/client/src/main/java/io/pravega/client/stream/notifications

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
fpj: @shrids That's an option, but why not have a `toString` method for metrics and logging purposes? It a pretty natural way of getting it.

shrids: Sure, will fix it as part of https://github.com/pravega/pravega/issues/2490

vijikarthi:
@EronWright do you want to take a look or can we merge it?

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +258,150 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+            builder.append("}");
+            return builder.toString();
+        }
+    }
+
+    /**
+     * register reader group metrics
+     *
+     */
+    private void registerMetrics() {
```
EronWright: I am unsure about the threading model for Flink metrics and for the reader group object.   Obviously the metrics are pulled using a different thread than the `run` thread.   Anything to be concerned about?

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -44,6 +52,26 @@
 
     private static final long serialVersionUID = 1L;
 
+    // ----- metrics field constants -----
+
+    private static final String PRAVEGA_READER_METRICS_GROUP = "PravegaReader";
+
+    private static final String READER_GROUP_METRICS_GROUP = "readerGroup";
```
EronWright: Seems that the various metrics here are all 'group-level', e.g. the unread bytes for the entire group.   However, Flink will report these metrics on a per-task basis.   I suppose that all the instances will be showing the same values (since they are all part of the same reader group).   I wonder whether we are missing a opportunity to report true instance-level metrics, e.g. how many segments are assigned to _this_ instance vs another.  Food for thought.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +258,150 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+        private final ReaderGroup readerGroup;
+        private final String separator = ",";
+
+        public StreamNamesGauge(ReaderGroup readerGroup) {
+            this.readerGroup = readerGroup;
+        }
+
+        @Override
+        public String getValue() {
+            return readerGroup.getStreamNames().stream().collect(Collectors.joining(separator));
```
EronWright: A key benefit of this gauge is to support programmatic discovery of which streams a job is working on.   I feel that we should use fully-qualified names here, e.g. `scope1/stream1`.

src/main/java/io/pravega/connectors/flink/AbstractReaderBuilder.java
```diff
@@ -137,6 +139,24 @@ protected PravegaConfig getPravegaConfig() {
                 .collect(Collectors.toList());
     }
 
+    /**
+     * enable/disable pravega reader metrics (default: enabled).
+     *
+     * @param enable boolean
+     * @return A builder to configure and create a reader.
+     */
+    public B enableMetrics(boolean enable) {
```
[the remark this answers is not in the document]
vijikarthi: Kafka connector allows the metrics to be configurable and I followed the same here. I don't see any harm in keeping the option since metric is enabled by default?

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +258,150 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+            return builder.toString();
+        }
+    }
+
+    /**
+     * register reader group metrics
+     *
+     */
+    private void registerMetrics() {
+        ReaderGroup readerGroup = createReaderGroup();
```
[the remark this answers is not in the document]
vijikarthi: Are you suggesting to have the reader extend from `AbstractRichFunction` to make use of `open()` and `close()` lifecycle? I think we need to make the `ReaderGroup` serializable to hold on to the instance and close() it properly as part of the lifecycle.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +258,150 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+        readerGroupMetricGroup.gauge(UNREAD_BYTES_METRICS_GAUGE, new UnreadBytesGauge(readerGroup));
+        readerGroupMetricGroup.gauge(READER_GROUP_NAME_METRICS_GAUGE, new ReaderGroupNameGauge(readerGroup));
+        readerGroupMetricGroup.gauge(SCOPE_NAME_METRICS_GAUGE, new ScopeNameGauge(readerGroup));
+        readerGroupMetricGroup.gauge(ONLINE_READERS_METRICS_GAUGE, new OnlineReadersGauge(readerGroup));
+        readerGroupMetricGroup.gauge(STREAM_NAMES_METRICS_GAUGE, new StreamNamesGauge(readerGroup));
+
+        Map<Stream, StreamCut> streamCuts = readerGroup.getStreamCuts();
+        for (Map.Entry<Stream, StreamCut> entry: streamCuts.entrySet()) {
+            Stream stream = entry.getKey();
+            MetricGroup streamMetricGroup = readerGroupMetricGroup.addGroup(STREAM_METRICS_GROUP + "." + stream.getStreamName());
```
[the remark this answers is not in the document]
vijikarthi: Will configure the group name like below
`STREAM_METRICS_GROUP + "." + stream.getScope()+ "_"+ stream.getStreamName()`

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +258,150 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+        @Override
+        public String getValue() {
+            return readerGroup.getStreamNames().stream().collect(Collectors.joining(separator));
+        }
+    }
+
+    /**
+     * Gauge for getting position information of each segment of a stream from reader group.
+     */
+    private static class SegmentPositionsGauge implements Gauge<String> {
```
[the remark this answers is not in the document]
vijikarthi: I haven't tested for multiple streams but I realized that I am filtering only on the stream name instead of stream and scope. Will update the code and test with multiple streams.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
vijikarthi: Will include scoped stream values

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
shrids: @fpj @vijikarthi 
The current implementation of StreamCut already implements a `toString()` method.
i.e. if we have `StreamCut sc` which has two segment with ` segmentNumber/offset : 0/ 10  and 1/10`
then `sc.toString()` will generate a string  of format `scope/stream:0=10, 1=20` I believe this is sufficient for logging purposes.

Issue #2490 will ensure that there is a way of representing StreamCut  in Base64 / String which can be easily passed as a program parameter ; it also supports ability to construct a StreamCut back from this textual representation.

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderTest.java
```diff
@@ -153,8 +153,9 @@ public void testCancellation() throws Exception {
     private static TestableFlinkPravegaReader<Integer> createReader() {
         ClientConfig clientConfig = ClientConfig.builder().build();
         ReaderGroupConfig rgConfig = ReaderGroupConfig.builder().stream(SAMPLE_STREAM).build();
+        boolean enableMetrics = true;
```
EronWright: Would be nice to see some actual assertions about the metrics being installed, but to obtain the metric group (and thus to make assertions about its contents), one needs a handle on the `operator` (which has `getMetricGroup()`).   A simple fix would be to amend `StreamSourceOperatorTestHarness` (which is our class) to have `getMetricGroup` (since the harness  has an `operator` field).

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -167,6 +184,32 @@ public void close() throws Exception {
         }
     }
 
+    /**
+     * Gauge for getting the scoped name information.
```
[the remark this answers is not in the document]
fpj: I think the use of scoped name here is not reflecting well what this metric wants to capture. At the same time, I'm not sure what you are referring as weird with respect scoped name; is it the use in Pravega or the use here specifically?

@vijikarthi could you in any case please clarify what this particular metric is trying to achieve?

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -187,11 +187,11 @@ public void close() throws Exception {
     /**
      * Gauge for getting the scoped name information.
```
fpj: Since you have changed from "scoped name" to "stream name" (which I think is better), shouldn't you also change in this comment block and in the variable definition inside the class for clarity?

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +274,166 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+        private final String stream;
+
+        public SegmentPositionsGauge(ReaderGroup readerGroup, String scope, String stream) {
+            this.readerGroup = readerGroup;
+            this.scope = scope;
+            this.stream = stream;
+        }
+
+        @Override
+        public String getValue() {
```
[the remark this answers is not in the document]
fpj: What is the consequence of a format change for metrics? I think it is only that the data is presented in a different way, and if I'm right there is no harm in having this in and changing later when we do it in Pravega.

src/main/java/io/pravega/connectors/flink/AbstractReaderBuilder.java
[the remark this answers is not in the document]
fpj: for another reference, this is what we do in Pravega atm:

https://github.com/pravega/pravega/blob/master/config/config.properties#L135

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +274,166 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+            return readerGroup.getGroupName();
+        }
+    }
+
+    /**
+     * Gauge for getting the scope name information from reader group.
+     */
+    private static class ScopeNameGauge implements Gauge<String> {
```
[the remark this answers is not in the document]
vijikarthi: I have kept the scope name metrics separate since a reader group scope could be different from the stream scope (multiple streams). We also report qualified stream name as another metric `StreamNamesGauge` which lists all the stream names as comma separated string.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
vijikarthi: From what I have tested so far, I did not see any issues in metrics reporting by Flink.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -220,6 +258,150 @@ private void triggerCheckpoint(String checkpointIdentifier) throws FlinkExceptio
...
+        ReaderGroup readerGroup = createReaderGroup();
+        MetricGroup pravegaReaderMetricGroup = getRuntimeContext().getMetricGroup().addGroup(PRAVEGA_READER_METRICS_GROUP);
+        MetricGroup readerGroupMetricGroup = pravegaReaderMetricGroup.addGroup(READER_GROUP_METRICS_GROUP);
+        readerGroupMetricGroup.gauge(UNREAD_BYTES_METRICS_GAUGE, new UnreadBytesGauge(readerGroup));
+        readerGroupMetricGroup.gauge(READER_GROUP_NAME_METRICS_GAUGE, new ReaderGroupNameGauge(readerGroup));
+        readerGroupMetricGroup.gauge(SCOPE_NAME_METRICS_GAUGE, new ScopeNameGauge(readerGroup));
+        readerGroupMetricGroup.gauge(ONLINE_READERS_METRICS_GAUGE, new OnlineReadersGauge(readerGroup));
+        readerGroupMetricGroup.gauge(STREAM_NAMES_METRICS_GAUGE, new StreamNamesGauge(readerGroup));
+
+        Map<Stream, StreamCut> streamCuts = readerGroup.getStreamCuts();
```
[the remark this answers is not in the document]
vijikarthi: `getStreamCuts()` is used only to fetch the scope and stream information to create appropriate metric groups since the `getStreamNames()` API does not return fully qualified stream details (https://github.com/pravega/pravega/issues/2860).

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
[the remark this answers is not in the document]
vijikarthi: This metric is to capture the fully qualified stream name (scope/stream) information of the `FlinkPravegaWriter` instance.

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderTest.java
vijikarthi: I have added a test case to assert the metric group being created

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
vijikarthi: I have updated the comment and the variable name

Issue #146, opened 2018-05-25
Support timestamp rows with FlinkPravegaTableSource


EronWright:
**Problem description**
@claudiofahey reports that there's no clear way to designate a particular column as timestamp-bearing, for the purposes of timestamp assignment / watermark generation.   Investigate whether there's generic support in the Table API or what must minimally be done.   Also see [`Kafka010JsonTableSource::setProctimeAttribute`](https://ci.apache.org/projects/flink/flink-docs-release-1.4/api/java/org/apache/flink/streaming/connectors/kafka/Kafka010JsonTableSource.html#setProctimeAttribute-java.lang.String-).

**Problem location**

**Suggestions for an improvement**

vijikarthi:
closed https://github.com/pravega/flink-connectors/pull/177

PR #177, opened 2018-08-05
[issue-146] Support timestamp rows with FlinkPravegaTableSource

vijikarthi:

**Change log description**
added processing and event time support to table source connector

**Purpose of the change**
To address the issue https://github.com/pravega/flink-connectors/issues/146

**What the code does**
Implements `DefinedProctimeAttribute` and `DefinedRowtimeAttributes` and enables API to supply the processing and event time capabilities to the table source connector

**How to verify it**
added integration test `FlinkPravegaTableITCase` to verify the implementation.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
@@ -188,7 +188,7 @@ public abstract class FlinkPravegaTableSource implements StreamTableSource<Row>,
 
         /**
          * Configures a field of the table to be a processing time attribute.
-         * The configured field must be present in the tabel schema and of type {@link Types#SQL_TIMESTAMP()}.
+         * The configured field must be present in the table schema and of type {@link Types#SQL_TIMESTAMP()}.
          *
          * @param proctimeAttribute The name of the processing time attribute in the table schema.
          * @return The builder.
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
@@ -19,12 +19,21 @@ import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.api.common.serialization.DeserializationSchema;
 import org.apache.flink.table.api.TableSchema;
+import org.apache.flink.table.api.Types;
+import org.apache.flink.table.api.ValidationException;
 import org.apache.flink.table.sources.BatchTableSource;
+import org.apache.flink.table.sources.DefinedProctimeAttribute;
+import org.apache.flink.table.sources.DefinedRowtimeAttributes;
+import org.apache.flink.table.sources.RowtimeAttributeDescriptor;
 import org.apache.flink.table.sources.StreamTableSource;
 import org.apache.flink.table.sources.TableSource;
+import org.apache.flink.table.sources.tsextractors.TimestampExtractor;
+import org.apache.flink.table.sources.wmstrategies.WatermarkStrategy;
 import org.apache.flink.types.Row;
 import org.apache.flink.util.Preconditions;
+import scala.Option;
 
+import java.util.Collections;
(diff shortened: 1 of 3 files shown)


src/test/java/io/pravega/connectors/flink/FlinkPravegaTableITCase.java
```diff
@@ -0,0 +1,345 @@
...
+            }
+        }
+
+        log.info("results: {}", RESULTS);
+
+        boolean compare = compare(RESULTS, getExpectedResultsRetracted());
+        assertTrue("Output does not match expected result", compare);
+    }
+
+    public void tesTableSourceBatch(FlinkPravegaJsonTableSource source) throws Exception {
```
claudiofahey: Typo in function name.

src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
```diff
@@ -119,6 +186,48 @@ public B withSchema(TableSchema schema) {
             return builder();
         }
 
+        /**
+         * Configures a field of the table to be a processing time attribute.
+         * The configured field must be present in the tabel schema and of type {@link Types#SQL_TIMESTAMP()}.
```
fpj: Typo in `table`.

src/test/java/io/pravega/connectors/flink/FlinkPravegaTableITCase.java
```diff
@@ -0,0 +1,345 @@
...
+                .field("user", Types.STRING())
+                .field("uri", Types.STRING())
+                .field("accessTime", Types.SQL_TIMESTAMP())
+                .build();
+
+        FlinkPravegaJsonTableSource source = FlinkPravegaJsonTableSource.builder()
+                                                .forStream(stream)
+                                                .withPravegaConfig(pravegaConfig)
+                                                .failOnMissingField(true)
+                                                //.withProctimeAttribute("accessTime")
```
fpj: Why is this commented out?

src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
```diff
@@ -35,7 +44,8 @@
  *
  * Supports both stream and batch environments.
  */
-public abstract class FlinkPravegaTableSource implements StreamTableSource<Row>, BatchTableSource<Row> {
+public abstract class FlinkPravegaTableSource implements StreamTableSource<Row>, BatchTableSource<Row>,
```
fpj: Is it worth documenting this feature?

src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
vijikarthi: fixed

src/test/java/io/pravega/connectors/flink/FlinkPravegaTableITCase.java
vijikarthi: We could use either `withProctimeAttribute` or `withRowtimeAttribute` to act on the time information. I have left the commented code there just to indicate the process time attribute usage.

src/main/java/io/pravega/connectors/flink/FlinkPravegaTableSource.java
vijikarthi: The code has inline comments and I believe it should be suffice.

src/test/java/io/pravega/connectors/flink/FlinkPravegaTableITCase.java
fpj: Would it make sense to have a test case for each? Also, what's the expected behavior in the case the user accidentally uses both.

vijikarthi: supplying both `proc` and `row` time attributes will throw an exception and hence cannot be used together. Having a test case for `proc` time is not required since it is based on the wall clock time and will have no bearing on the record timestamp.
> "org.apache.flink.table.api.ValidationException: Field 'accessTime' must not be processing time and rowtime attribute at the same time."

Issue #201, opened 2018-12-10
Update connector to support Pravega client API refactor


vijikarthi:
**Problem description**
With the recent changes to Pravega client API [see here](https://github.com/pravega/pravega/pull/3126), flink connector build is failing. Make changes to the connector to support Pravega client API refactor changes.

**Problem location**
Flink connector

**Suggestions for an improvement**

PR #202, opened 2018-12-10
[issue-201] Update connector to support Pravega client API refactor

vijikarthi:

**Change log description**
Updated connector code to fix the deprecated API build error because of the recent changes introduced in Pravega client API.

**Purpose of the change**
To address the issue https://github.com/pravega/flink-connectors/issues/201

**What the code does**
- Updated Pravega artifact and submodule version to latest snapshot
- Addressed the Pravega client API refactoring changes. 
- Fixed unit/integration tests.

**How to verify it**

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -25,7 +25,7 @@ gradleMkdocsPluginVersion=1.1.0
 
 # Version and base tags can be overridden at build time.
 connectorVersion=0.4.0-SNAPSHOT
-pravegaVersion=0.4.0-50.da91e55-SNAPSHOT
+pravegaVersion=0.5.0-49.b80b627-SNAPSHOT
 apacheCommonsVersion=3.7
 
 # flag to indicate if Pravega sub-module should be used instead of the version defined in 'pravegaVersion'
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
@@ -11,13 +11,15 @@ package io.pravega.connectors.flink;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.pravega.client.ClientConfig;
-import io.pravega.client.ClientFactory;
+import io.pravega.client.EventStreamClientFactory;
 import io.pravega.client.stream.EventStreamWriter;
+import io.pravega.client.stream.TransactionalEventStreamWriter;
 import io.pravega.client.stream.EventWriterConfig;
 import io.pravega.client.stream.Serializer;
 import io.pravega.client.stream.Stream;
 import io.pravega.client.stream.Transaction;
 import io.pravega.common.Exceptions;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.flink.annotation.VisibleForTesting;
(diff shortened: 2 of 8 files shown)


gradle.properties
```diff
@@ -25,7 +25,7 @@ gradleMkdocsPluginVersion=1.1.0
 
 # Version and base tags can be overridden at build time.
 connectorVersion=0.4.0-SNAPSHOT
-pravegaVersion=0.4.0-50.da91e55-SNAPSHOT
+pravegaVersion=0.5.0-49.b80b627-SNAPSHOT
```
fpj: I'm confused, why are we moving to `master`? Keep in mind that the API refactoring was merged to `master` (`0.5.0-SNAPSHOT`), not `r0.4`.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -354,22 +356,35 @@ public String toString() {
     @VisibleForTesting
     abstract class AbstractInternalWriter {
 
-        protected final EventStreamWriter<T> pravegaWriter;
+        @Getter
+        private EventStreamWriter<T> pravegaWriter;
 
-        AbstractInternalWriter(ClientFactory clientFactory) {
+        @Getter
+        private TransactionalEventStreamWriter<T> pravegaTxnWriter;
```
kevinhan88: Since this is AbstractInternalWriter, why not make it real abstract? such as define only one variable "streamWriter" as AutoCloseable, plus keep the switch of txnWriter (better name it as ifTxnWriter).
The subclass can cast the returned streamWriter appropriately based on ifTxnWriter.

The reason of this suggested refactoring is: object existence test as txn writer flag might not be always reliable. Say if the incoming config is invalid then no writer can be created, but there's no handling of such scenario in constructor, then the object test might be misleading.

gradle.properties
vijikarthi: I believe the issue is noticed with the connector/master code tested against the latest Pravega/master.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
vijikarthi: There is no common interface that can be returned and that's the reason for keeping two separate references in the base class. The child class know which one they are operating on and will not get in to the issue that you are referring to.

kevinhan88: This sounds a bottom up usage instead of top-down design, then the code looks fine to me

gradle.properties
fpj: Ok, to confirm, we want to move flink connector master to pravega master? we might want to create an `r0.4` branch first, no?

fpj:
@vijikarthi to confirm, we will wait until we create the `r0.4` before we merge this one, yes?

vijikarthi:
> @vijikarthi to confirm, we will wait until we create the `r0.4` before we merge this one, yes?

I have created `r0.4` branch in connectors repository and will update the gradle properties to `0.4.0` release version (both connector and pravega) after we release Pravega `0.4.0`. In the meanwhile, to this PR, I have advanced the connector version to `0.5.0-SNAPSHOT`

Issue #214, opened 2019-02-25
Enable spotbugs


fpj:
**Problem description**
We need to enable spotbugs for static analysis:

https://spotbugs.github.io/

**Problem location**
Gradle build.

**Suggestions for an improvement**
Add to gradle build, check how we do it in Pravega.

vijikarthi:
closed https://github.com/pravega/flink-connectors/pull/218

PR #218, opened 2019-03-08
[issue-214] Integrate spotbugs into flink connector build

elizabethbain:
Signed-off-by: Elizabeth Bain <elizabeth.bain@emc.com>

**Change log description**
Install spotbugs analysis into flink connector build

**Purpose of the change**
Use latest static analysis tool for potential bug checking. This fix is to address the issue https://github.com/pravega/flink-connectors/issues/214

**What the code does**
Looks for specified patterns in the code base to identify possible bug issues
Replaces prior findbugs tool for static analysis
Modified build.gradle to include spot bugs specifications and remove prior find bugs references
Add spotbugs gradle file and checkstyle files with specifications on patterns to look for and remove findbugs files
upgrade gradle to 4.10.2 from 4.5 - needed to fix spotbugs reporting error

**How to verify it**
check build to see spotbugs tasks ran: spotbugsMain and spotbugsTest


src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java
```diff
@@ -40,7 +41,8 @@ public SegmentRange getSegmentRange() {
     }
 
     // --------------------------------------------------------------------
-
+    // cconstructor guards segment range from being null
+    @SuppressFBWarnings("NP_NULL_ON_SOME_PATH_MIGHT_BE_INFEASIBLE")
```
vijikarthi: Why the warning is suppressed?

elizabethbain: There was only one constructor and it looked like it guarded against nulls there. I debated on putting the checks in the equals method or  relying on the constructor constraints and putting the suppress in. I'm equally fine with either way. In hindsight, it's possible to change it after construction. I can put the guards  in the method instead.

build.gradle
```diff
@@ -12,31 +12,49 @@
...
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
     }
     dependencies {
         classpath group: 'com.github.jengelman.gradle.plugins', name:'shadow', version: shadowGradlePlugin
         classpath "com.jfrog.bintray.gradle:gradle-bintray-plugin:${bintrayPluginVersion}"
         classpath group: 'org.hidetake', name: 'gradle-ssh-plugin', version: gradleSshPluginVersion
         classpath "ru.vyarus:gradle-mkdocs-plugin:${gradleMkdocsPluginVersion}"
+        classpath "gradle.plugin.com.github.spotbugs:spotbugs-gradle-plugin:1.6.9"
```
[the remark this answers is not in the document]
elizabethbain: Sure.  I believe it's already in the properties. I just need to use it.

build.gradle
```diff
@@ -12,31 +12,49 @@
...
     }
 }
 
+plugins {
+  id "com.github.spotbugs" version "1.6.9"
+
```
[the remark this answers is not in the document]
elizabethbain: Sure.  I believe it's already in the properties. I just need to use it.

gradlew.bat
```diff
@@ -1,22 +1,10 @@
...
-@rem
-@rem  Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
-@rem
-@rem  Licensed under the Apache License, Version 2.0 (the "License");
-@rem  you may not use this file except in compliance with the License.
-@rem  You may obtain a copy of the License at
-@rem
-@rem      http://www.apache.org/licenses/LICENSE-2.0
-@rem
-
```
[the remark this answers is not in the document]
elizabethbain: I upgraded gradle version through gradlew.  It must have removed the headers. I'll restore them.

gradlew
```diff
@@ -1,21 +1,11 @@
...
-#
-# Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-
```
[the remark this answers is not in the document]
elizabethbain: I upgraded gradle version through gradlew.  It must have removed the headers. I'll restore them.

build.gradle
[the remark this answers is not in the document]
elizabethbain: Done

build.gradle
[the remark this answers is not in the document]
elizabethbain: This was not as easy as I would have hoped. Found that there was an issue with gradle structure for plugins ids as shown in this gradle issue:
https://github.com/gradle/gradle/issues/1697
  It required adding a block to settings. gradle as shown at the bottom of the issue description in the link.  Making those changes to settings enabled me to get past the issue.

gradlew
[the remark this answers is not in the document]
elizabethbain: restored the header.

gradlew.bat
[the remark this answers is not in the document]
elizabethbain: restored the header

src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java
elizabethbain: Added the guard into the source code and removed the suppression annotation

src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java
```diff
@@ -66,22 +64,30 @@ public boolean equals(Object o) {
...
+                    segmentRange.getStartOffset() == that.getSegmentRange().getStartOffset() &&
+                    segmentRange.getEndOffset() == that.getSegmentRange().getEndOffset() &&
+                    segmentRange.getSegmentId() == that.getSegmentRange().getSegmentId();
         }
 
-        return splitId == that.splitId &&
-                segmentRange.getStartOffset() == that.getSegmentRange().getStartOffset() &&
-                segmentRange.getEndOffset() == that.getSegmentRange().getEndOffset() &&
-                segmentRange.getSegmentId() == that.getSegmentRange().getSegmentId();
+        // based on other checks earlier -
```
vijikarthi: Do we need to keep these comments?

src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java
```diff
@@ -66,22 +64,30 @@ public boolean equals(Object o) {
...
-        String thatScope = that.getSegmentRange().getScope();
-        if (thisScope == null ? thatScope != null : !thisScope.equals(thatScope)) {
-            return false;
-        }
-
-        String thisStream = segmentRange.getStreamName();
-        String thatStream = that.getSegmentRange().getStreamName();
-        if (thisStream == null ? thatStream != null : !thisStream.equals(thatStream)) {
-            return false;
+        if (segmentRange != null) {
```
vijikarthi: do we have unit test that validates the equals() method for various test conditions? If not, could you please add?

checkstyle/spotbugs-include.xml
```diff
@@ -48,7 +46,7 @@
     <!-- Experimental and not fully vetted bug patterns -->
     <Match>
         <Bug category="EXPERIMENTAL" />
-        <Rank value="9"/>
+        <Confidence value="1"/>
```
vijikarthi: Is confidence equivalent of rank? what does value 1 means?

gradlew
```diff
@@ -175,7 +175,7 @@ APP_ARGS=$(save "$@")
 eval set -- $DEFAULT_JVM_OPTS $JAVA_OPTS $GRADLE_OPTS "\"-Dorg.gradle.appname=$APP_BASE_NAME\"" -classpath "\"$CLASSPATH\"" org.gradle.wrapper.GradleWrapperMain "$APP_ARGS"
 
 # by default we should be in the correct project dir, but when run from Finder on Mac, the cwd is wrong
-if [[ "$(uname)" == "Darwin" ]] && [[ "$HOME" == "$PWD" ]]; then
+if [ "$(uname)" = "Darwin" ] && [ "$HOME" = "$PWD" ]; then
```
vijikarthi: why is the equality check changed from `==` to `=`?

src/test/java/io/pravega/connectors/flink/PravegaInputSplitTest.java
```diff
@@ -0,0 +1,73 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.pravega.connectors.flink;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
vijikarthi: Please change the bulk import style to specific class import.

src/test/java/io/pravega/connectors/flink/PravegaInputSplitTest.java
```diff
@@ -0,0 +1,73 @@
...
+        // compare one to itself
+        assertTrue(split1.equals(split1));
+
+        // compare one to the other
+        assertTrue(split1.equals(split2));
+
+    }
+
+    @Test
+    public void testNotEquals() {
```
vijikarthi: can you also make sure to do null checks validation? 
range1 == range2 == null
range1 == null && range1 != null
range1 != null && range2 == null

checkstyle/spotbugs-include.xml
elizabethbain: No confidence used to be priority in prior versions. It indicates a confidence in the possibility that spotbugs is correct for that bug type. It indicates the confidence that the found bug is actually a bug. FindBugs employs heuristics which may go wrong. It varies from 1 (highest confidence) to 5 (lowest confidence, to be disregarded). 

Rank indicates how critical a bug is or in spotbugs terms how "scary" it is. It indicate the impact of that bug on the system - how severe it would be.  The values range this way:
Scariest (1-4),
Scary (5-9),
Troubling (10-14)
Of Concern (rank 15-20)
One analog is; So, a certain NullPointerException might get a 1, and a potential performance improvement might just get a 12.

The values match those in Pravega spot bugs. The expected time for me to do this was evidently short and Tom had converted the Pravega build, so I trusted his judgement, since I didn't have the leeway to investigate each of the potential spotbugs type.

gradlew
elizabethbain: This was a side effect of the upgrade to a newer gradle version. When I tried to put it back to the old equality expression it gave me a bizarre whining about that line.
./gradlew: 178: [: Linux: unexpected operator

 So I checked on the pravega main project gradlew and it also was like the upgraded one  - 
if [ "$(uname)" = "Darwin" ] && [ "$HOME" = "$PWD" ]; then

src/test/java/io/pravega/connectors/flink/PravegaInputSplitTest.java
elizabethbain: I don't think range1 or range2 can be null. The constructor for split input has the following precondition check which makes it impossible to create one with a null range in test:
Preconditions.checkNotNull(segmentRange, "segmentRange");
SegmentRange is an interface with SegmentRangleImpl having the following annotations:
@EqualsAndHashCode
So I believe those are covered by cases that check segmentRange.equals(segmentRange). since both null should be true and one null and one not should yield false. 
I tried doing these cases with Mokito with mock ranges and when (...).thenReturn(null), which worked for 2 of them but not the third. Its been awhile but I think it was the middle one where it couldn't appropriately resolve one of the mocks. If you can come up with a scheme that will work, I'd be happy to implement it.

checkstyle/spotbugs-exclude.xml
```diff
@@ -8,7 +8,7 @@
     <Match> <!-- No practical consequence and generate a lot of warnings -->
```
vijikarthi: Looks like both the files (include and exclude.xml) don't have the copyright headers. Could you please add?

settings.gradle
```diff
@@ -8,8 +8,20 @@
  *     http://www.apache.org/licenses/LICENSE-2.0
  *
  */
+pluginManagement {
+  it.resolutionStrategy {
+      it.eachPlugin {
+        if (it.requested.id.id == 'com.github.spotbugs') {
+          it.useVersion("${spotbugsPluginVersion}")
```
vijikarthi: requires one extra space

src/main/java/io/pravega/connectors/flink/PravegaInputSplit.java
```diff
@@ -40,7 +40,7 @@ public SegmentRange getSegmentRange() {
     }
 
     // --------------------------------------------------------------------
-
+    // cconstructor guards segment range from being null
```
vijikarthi: typo

build.gradle
```diff
@@ -12,31 +12,46 @@
...
         classpath "com.jfrog.bintray.gradle:gradle-bintray-plugin:${bintrayPluginVersion}"
         classpath group: 'org.hidetake', name: 'gradle-ssh-plugin', version: gradleSshPluginVersion
         classpath "ru.vyarus:gradle-mkdocs-plugin:${gradleMkdocsPluginVersion}"
+        classpath "gradle.plugin.com.github.spotbugs:spotbugs-gradle-plugin:${spotbugsPluginVersion}"
     }
 }
-
+    plugins {
+       id "com.github.spotbugs" 
+    }
```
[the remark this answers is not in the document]
elizabethbain: Hi, 
   When I used that convention it wouldn't recognize or run spotbug. Pravega has that convention but Tom mentioned they had to convert build.gradle to use allProjects to get spotbugs to work. I had hoped the upgrade of gradle would fix this too, but I just tried it and spotbugs didn't run (and it didn't generate an error - the advantage is it also didn't generate all the ugly warnings.). I didn't feel like I knew enough to refactor build.gradle to use  projects or if it even made sense, so I had found this format as a reference and it made spotbugs start working. I can move it below the other apply plugin lines so it doesn't stick out so much....

PR #274, opened 2019-09-20
[issue-265] Reader side: Implement watermark with Pravega

crazyzhou:
**Change log description**

- Fix build error caused by breaking API change in `pravega-client`
- Implement reader with Pravega watermark. 
- Add watermark support in Stream Table Source of Table API. 
- Check only 1 stream linked with 1 `FlinkPravegaReader` when using `AssignerWithTimeWindow`.
- Add three test cases to make sure it works

**Purpose of the change**

- Support Pravega watermark in reader/source
- Merge into Flink `1.7` branch first because I start from it. Will cherry pick into `1.8` and `master` later

**What the code does**

- Implement Pravega watermark in both normal and Table API streaming reader/source
- We offer an `AssignerWithTimeWindow` interface for users. Users can use `withTimestampAndWatermark` to use Pravega watermark in the source function builder.
- We add watermark support `AssignerWithTimeWindow` class in the connectorProperties and integrate watermark in the underlying DataStreamSource of the Stream Table Source. Users can use `timestampFromSource` and `watermarkFromSource` in the `Rowtime` to use Pravega watermark.

**Limitations**

Due to the serialization validation for `connectorProperties`, only `public` `static-inner/outer` class implements `AssignerWithTimeWindow` is supported as a parameter of `withTimestampAndWatermark` in Table API stream table source. See: https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/util/InstantiationUtil.java#L489

**How to verify it**

Add three test cases:
1. TestHarness test for `FlinkPravegaReader`, watermark can periodically generated as expected.
2. Integration test for `FlinkPravegaReader`, guarantee no data loss and duplicate with windowing and watermark.
3. Interface validation test for Pravega `StreamTableSource` with watermark, validation passes.

`./gradlew clean build` pass.


src/main/java/io/pravega/connectors/flink/watermark/LowerBoundAssigner.java
```diff
@@ -0,0 +1,34 @@
...
+    @Override
+    public abstract long extractTimestamp(T element, long previousElementTimestamp);
+
+    // built-in watermark implementation which emits the lower bound - 1
+    @Override
+    public Watermark getWatermark(TimeWindow timeWindow) {
+        if (timeWindow == null || timeWindow.isNearHeadOfStream()) {
+            return null;
+        }
+        return new Watermark(timeWindow.getLowerTimeBound() - 1L);
```
EronWright: Is there a possible underflow here?   Would `getLowerTimeBound` ever be equal to `Long.MIN_VALUE`?

src/main/java/io/pravega/connectors/flink/watermark/LowerBoundAssigner.java
```diff
@@ -0,0 +1,34 @@
...
+    public LowerBoundAssigner() {
+    }
+
+    @Override
+    public abstract long extractTimestamp(T element, long previousElementTimestamp);
+
+    // built-in watermark implementation which emits the lower bound - 1
+    @Override
+    public Watermark getWatermark(TimeWindow timeWindow) {
+        if (timeWindow == null || timeWindow.isNearHeadOfStream()) {
```
EronWright: Could you add a little commentary in the source code about why `isNearHeadOfStream` is used here?  I don't understand it.

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderITCase.java
```diff
@@ -78,6 +84,75 @@ public void testMultipleSourcesMultipleSegments() throws Exception {
...
+            // if these calls complete without exception, then the test passes
+            try {
+                env.execute();
+            } catch (Exception e) {
+                if (!(ExceptionUtils.getRootCause(e) instanceof SuccessException)) {
+                    throw e;
+                }
+            }
+
+            final long executeEnd = System.nanoTime();
```
EronWright: Where are the assertions about the correctness ?

src/test/java/io/pravega/connectors/flink/FlinkPravegaTableSourceTest.java
```diff
@@ -171,6 +174,51 @@ public void testTableSourceDescriptor() {
...
+
+        Stream stream = Stream.of(scopeName, streamName);
+        PravegaConfig pravegaConfig = PravegaConfig.fromDefaults()
+                .withControllerURI(URI.create(controllerUri))
+                .withDefaultScope(scopeName);
+
+        // construct table source using descriptors and table source factory
+        Pravega pravega = new Pravega();
+        pravega.tableSourceReaderBuilder()
+                .withTimestampAndWatermark(new MyAssigner())
```
EronWright: I have a concern about the table API, could you provide some links about how the Kafka table source does it?   One specific concern is whether the `AssignerWithTimeWindows` is an appropriate interface for this situation.  I recall there's `rowtime` stuff which tells the table API about which column contains the event time.

crazyzhou: Yes, Flink has a `rowtime` which tells the timestamp and watermark. However, this only support watermark of `periodicTimestampAssigner` or `punctuatedTimestampAssigner` as its strategy, See: https://github.com/apache/flink/blob/master/flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/wmstrategies/WatermarkStrategy.java#L33. Both of them are data-driven watermark and it is totally different from our Pravega's watermark. It is a good news that it supports underlying `dataStreamSource` to have its own timestamp and watermark, so I choose to implement it in the underlying stream. User can use ````

crazyzhou: To answer your first question, Kafka did nothing to support its own watermark strategy in Table API. It only put effort in their normal consumer at https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaConsumerBase.java#L293 . I can find nothing related to timestamp or watermark in its table API's properties, I think Flink treats Table API's row time attributes as data-driven mostly.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -368,9 +368,9 @@ public String toString() {
                     .transactionTimeoutTime(txnLeaseRenewalPeriod)
                     .build();
             if (txnWriter) {
-                pravegaTxnWriter = clientFactory.createTransactionalEventWriter(stream.getStreamName(), eventSerializer, writerConfig);
+                pravegaTxnWriter = clientFactory.createTransactionalEventWriter(name(), stream.getStreamName(), eventSerializer, writerConfig);
             } else {
-                pravegaWriter = clientFactory.createEventWriter(stream.getStreamName(), eventSerializer, writerConfig);
+                pravegaWriter = clientFactory.createEventWriter(name(), stream.getStreamName(), eventSerializer, writerConfig);
```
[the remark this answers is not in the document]
crazyzhou: Add this to let the build pass since its a breaking API change in `pravega-client`.

src/main/java/io/pravega/connectors/flink/watermark/LowerBoundAssigner.java
crazyzhou:
```
    /**
     * Returns true if the reader is currently near the head of the stream and therefor no lower time bound can be obtained.
     */
    public boolean isNearHeadOfStream() {
        return lowerTimeBound == null;
    }
```
This is a check if the lowerbound is null, we don't emit a watermark.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -488,18 +574,40 @@ protected ReaderGroupManager createReaderGroupManager() {
...
+        }
+
         @Override
         protected DeserializationSchema<T> getDeserializationSchema() {
             Preconditions.checkState(deserializationSchema != null, "Deserialization schema must not be null.");
             return deserializationSchema;
         }
 
+        @Override
+        protected SerializedValue<AssignerWithTimeWindows<T>> getAssignerWithTimeWindows() {
```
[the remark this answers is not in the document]
crazyzhou: I can accept this, but it might be confused with Flink's `org.apache.flink.streaming.api.functions.TimestampAssigner` which only contains the `extractTimestamp` method. Our `AssignerWithTimeWindows` extends that, so it has not only this method but also `getWatermark(TimeWindow)` method.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -183,6 +247,21 @@ public void run(SourceContext<T> ctx) throws Exception {
...
+            AssignerWithTimeWindows<T> assigner = null;
+            // If it is event time, register a watermark emitter
+            if (isEventTimeMode()) {
+                assigner = assignerWithTimeWindows.deserializeValue(getRuntimeContext().getUserCodeClassLoader());
+                PeriodicWatermarkEmitter periodicEmitter = new PeriodicWatermarkEmitter(
+                        pravegaReader,
+                        ctx,
+                        getRuntimeContext().getUserCodeClassLoader(),
+                        ((StreamingRuntimeContext) getRuntimeContext()).getProcessingTimeService(),
+                        getRuntimeContext().getExecutionConfig().getAutoWatermarkInterval());
```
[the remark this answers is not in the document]
crazyzhou: It is `0`: https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java#L125, I'll add it as an variable of the class to make it easier to use and check. I'd like to start another conversation here about this value. Since Pravega now updates the `TimeWindows` every 10 seconds, should we add more checks like ` interval > 10000ms` because any duration much shorter that 10 seconds is now actually useless as `TimeWindow` keeps the same.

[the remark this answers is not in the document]
maddisondavid: It looks like it gets changed to `200` as side effect of changing the Time Characteristic.

https://github.com/apache/flink/blob/7ea00ad9004f920e0c4c1b6ed906e14418951272/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/StreamExecutionEnvironment.java#L642-L649 but obviously we should still have some validation as a user can change this AFTER setting the Time Characteristic 

```
	public void setStreamTimeCharacteristic(TimeCharacteristic characteristic) {
		this.timeCharacteristic = Preconditions.checkNotNull(characteristic);
		if (characteristic == TimeCharacteristic.ProcessingTime) {
			getConfig().setAutoWatermarkInterval(0);
		} else {
			getConfig().setAutoWatermarkInterval(200);
		}
	}
```

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderITCase.java
crazyzhou: The `IntSequenceExactlyOnceValidator` guarantees that no data loss or duplicate.
The `ExceptionUtils.getRootCause(e) instanceof SuccessException` guarantees that no error in the processing pipeline.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -165,6 +180,55 @@ void initialize() {
...
         log.info("Creating reader group: {}/{} for the Flink job", this.readerGroupScope, this.readerGroupName);
         createReaderGroup();
+        if (isEventTimeMode()) {
+            Preconditions.checkArgument(readerGroup.getStreamNames().size() == 1,
+                    "Only 1 Pravega stream is allowed in the event-time mode");
+        }
+    }
+
+    private boolean isEventTimeMode() {
+        return assignerWithTimeWindows != null;
```
maddisondavid: It's unfortunate that there is no way to get the actual Time Characteristic of the environment.  The connector is making an ASSUMPTION that if the user doesn't specify an assigner then we're working in `EventTime` but it could be that the environment is working in `EventTime` but that they forgot to add the assigner.

It would be nice if there was a way to make this assertion that the connector and the stream are setup to work in the same time characteristic.

src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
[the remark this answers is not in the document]
crazyzhou: Created #276 for this single issue

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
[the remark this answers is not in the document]
crazyzhou: Thanks for correcting me. I added a precondition check during the `FlinkPravegaReader` `open` function. That's the first place where the `RuntimeContext` is being initialized and accessible.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
crazyzhou: There should be choice that user uses their own watermark by `assignTimestampAndWatermarks` operator and do not use Pravega watermarks. There is also a circumstance that the writer of the `stream` did not call `noteTime()`, we can't block the reader to use `EventTime` mode to read the stream. I think "no assertion" is reasonable here.

src/main/java/io/pravega/connectors/flink/Pravega.java
```diff
@@ -253,6 +280,24 @@ protected TableSourceReaderBuilder withDeserializationSchema(DeserializationSche
...
+         * @param assignerWithTimeWindows the timestamp and watermark assigner.
+         * @return TableSourceReaderBuilder instance.
+         */
+        // TODO: Due to the serialization validation for `connectorProperties`, only `public` `static-inner/outer` class implements
+        // `AssignerWithTimeWindow` is supported as a parameter of `withTimestampAssigner` in Table API stream table source.
+        protected TableSourceReaderBuilder withTimestampAssigner(AssignerWithTimeWindows<Row> assignerWithTimeWindows) {
+            try {
+                ClosureCleaner.clean(assignerWithTimeWindows, true);
+                this.assignerWithTimeWindows = new SerializedValue<>(assignerWithTimeWindows);
+            } catch (IOException e) {
```
maddisondavid: `ClosureClean` will throw an `InvalidProgramException` if the object fails serialization and not `IOException` (which is caught and transformed by `ClosureClean`.

`InvalidProgramException` does not extend `IOException` and therefore this will not catch the Exception

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderITCase.java
maddisondavid: Brian is correct, the `IntSequenceExactlyOnceValidator` is checking for correctness and either calling `Assert.fail` if there's a problem or throwing a `SuccessException` if all expected messages are received in the correct order.

That said, it's confusing if it DOES fail because the original `Assert.fail` message is lost with the reason WHY it failed.  I've now fixed this with an update that rethrows the original `AssertException` if that's the reason it failed.

src/main/java/io/pravega/connectors/flink/watermark/LowerBoundAssigner.java
maddisondavid: I've added a comment to LowerBoundAssigner

PR #301, opened 2019-12-04
[issue-298] Fix tests that transaction timeout should be greater than 10 seconds

### LINKED ISSUE #298: Force transaction timeout to be greater than 10 seconds
**Problem description**
Due to https://github.com/pravega/pravega/pull/4400 
we should fix tests where the transaction timeouts are less than 10 second.

**Problem location**
Tests

Change it to `30` seconds

crazyzhou:

**Change log description**
- Force transaction timeout to be greater than 10 seconds

**Purpose of the change**
Fixes #298 

**What the code does**
Change the default txn timeout in mock sink function from `30ms` to `30s` 

**How to verify it**

Target to master, should cherry-pick to all `0.7` and `0.6` branch


maddisondavid:
The title `Force transaction timeout to be greater than 10 seconds` makes it sound like this is a change to the connector. I just want to confirm that this is just fixing the tests to be inline with the new `0.6.1` requirements. (if that's the case we should update the title)

PR #305, opened 2020-01-06
[issue-180] Provide method for deserialization with metadata

### LINKED ISSUE #180: Expose the `EventRead` metadata along with a value
**Problem description**
Make it possible to create a `FlinkPravegaReader` that provides event metadata (e.g. `EventPosition`) for each event.    This would facilitate the development of jobs that care about the stream position of the event data, e.g. for indexing purposes.

Normally, the data stream that is produced is of type `DataStream<T>` to convey the deserialized event data of type `T`.   To also convey metadata implies a data stream of type `DataStream<EventRead<T>>`.   For example:

```java
FlinkPravegaReader<EventRead<Double>> reader = ...;
DataStream<EventRead<Double>> datastream = env.addSource(reader);
datastream.map(e -> {
  assert e.isEvent();
  LOG.info("event: data={}, position={}", e.getEvent(), e.getEventPosition());
});
```

The `EventRead` class might not be ideal; we could introduce a similar type based on discussion.

One complication is that the `DeserializationSchema<T>` that is provided to the reader is typed to the event data type.   Some cleverness in the builder would be needed to disambiguate the deserialized type  `<T>` from the data stream type `EventRead<T>`.    May also need to leverage Flink's `OutputTypeConfigurable`.

crazyzhou:

**Change log description**
- [issue-180] Provide method for deserialization with metadata

**Purpose of the change**
Fixes #180 

**What the code does**
Provide a method `deserializeWithMetadata` for user to override that can leverage `EventRead` in the deserialization.
Added unit test for the case.

**How to verify it**
Target to master, should cherry-pick to all `0.7` branch

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -20,6 +20,7 @@ import io.pravega.client.stream.ReaderGroup;
 import io.pravega.client.stream.ReaderGroupConfig;
 import io.pravega.client.stream.Stream;
 import io.pravega.client.stream.StreamCut;
+import io.pravega.connectors.flink.serialization.PravegaDeserializationSchema;
 import io.pravega.connectors.flink.watermark.AssignerWithTimeWindows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.flink.api.common.ExecutionConfig;
@@ -268,7 +269,10 @@ public class FlinkPravegaReader<T>
             // main work loop, which this task is running
             while (this.running) {
                 final EventRead<T> eventRead = pravegaReader.readNextEvent(eventReadTimeout.toMilliseconds());
-                final T event = eventRead.getEvent();
+
+                final T event = this.deserializationSchema instanceof PravegaDeserializationSchema ?
+                        ((PravegaDeserializationSchema<T>) deserializationSchema).deserializeWithMetadata(eventRead) :
+                        eventRead.getEvent();
 
                 // emit the event, if one was carried
                 if (event != null) {
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -267,8 +267,8 @@ public class FlinkPravegaReader<T>
                 periodicEmitter.start();
             }
 
-            Function<EventRead<T>, T> deserFunc = this.deserializationSchema instanceof PravegaDeserializationSchema ?
-                    ((PravegaDeserializationSchema<T>) deserializationSchema)::deserializeWithMetadata :
+            final Function<EventRead<T>, T> deserFunc = this.deserializationSchema instanceof PravegaDeserializationSchema ?
+                    ((PravegaDeserializationSchema<T>) deserializationSchema)::extractEvent :
                     (eventRead) -> eventRead.getEvent();
 
             // main work loop, which this task is running
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -47,6 +47,7 @@ import java.io.IOException;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static io.pravega.connectors.flink.util.FlinkPravegaUtils.createPravegaReader;
@@ -266,13 +267,15 @@ public class FlinkPravegaReader<T>
                 periodicEmitter.start();
             }
 
+            Function<EventRead<T>, T> deserFunc = this.deserializationSchema instanceof PravegaDeserializationSchema ?
+                    ((PravegaDeserializationSchema<T>) deserializationSchema)::deserializeWithMetadata :
+                    (eventRead) -> eventRead.getEvent();
+
             // main work loop, which this task is running
             while (this.running) {
                 final EventRead<T> eventRead = pravegaReader.readNextEvent(eventReadTimeout.toMilliseconds());
 
-                final T event = this.deserializationSchema instanceof PravegaDeserializationSchema ?
-                        ((PravegaDeserializationSchema<T>) deserializationSchema).deserializeWithMetadata(eventRead) :
-                        eventRead.getEvent();
+                final T event = deserFunc.apply(eventRead);
 
                 // emit the event, if one was carried
                 if (event != null) {
--- a/src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
+++ b/src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
@@ -31,7 +31,7 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  * exposes the produced type (TypeInformation) to allow Flink to configure its internal
  * serialization and persistence stack.
  *
- * <p>An additional method {@link #deserializeWithMetadata(EventRead)} is provided for
+ * <p>An additional method {@link #extractEvent(EventRead)} is provided for
  * applying the metadata in the deserialization. This method can be overriden in the extended class. </p>
  */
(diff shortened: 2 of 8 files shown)


src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -268,7 +269,10 @@ public void run(SourceContext<T> ctx) throws Exception {
             // main work loop, which this task is running
             while (this.running) {
                 final EventRead<T> eventRead = pravegaReader.readNextEvent(eventReadTimeout.toMilliseconds());
-                final T event = eventRead.getEvent();
+
+                final T event = this.deserializationSchema instanceof PravegaDeserializationSchema ?
```
EronWright: Consider doing the `instanceof` check once, outside the `while` loop, for better performance.   One approach would be like:
```
Function<EventRead,T> f = this.deserializationSchema instanceof PravegaDeserializationSchema ? this.deserializationSchema::deserializeWithMetadata : (eventRead) -> eventRead.getEvent();
```

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderTest.java
```diff
@@ -433,6 +499,30 @@ public boolean isEndOfStream(Integer nextElement) {
...
+        public TestDeserializationSchemaWithMetadata() {
+            super(Integer.class, new IntegerSerializer());
+        }
+
+        @Override
+        public Integer deserializeWithMetadata(EventRead<Integer> eventRead) {
+            if (isEndOfStream(eventRead.getEvent())) {
+                return eventRead.getEvent();
+            }
+            return eventRead.getEventPointer().toBytes().remaining();
```
claudiofahey: My impression is that this method should return an object that contains the event data (an Integer in this case) as well as the EventPointer. This appears to be ignoring the event data and only returning the size of the EventPointer. How is this useful? I recommend changing this test so that it uses the metadata in a useful way. This should provide a good integration test as well as an example for how an end-user can actually use it.

claudiofahey: Also, use JSON for serialization as it is more realistic.

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
crazyzhou: That's very helpful. I'll change it.

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderTest.java
crazyzhou: Yes, when I wrote these, I was also trying to think in this unit test, whether I should keep it simple(hold the integer output just to test functionality) or keep it interpretable(more like a guiding example). I can change that with your suggestion.

src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
```diff
@@ -126,6 +130,18 @@ public boolean isEndOfStream(T nextElement) {
...
 
+    /**
+     * An method for applying the metadata in deserialization.
+     * Override it in the custom extended {@link PravegaDeserializationSchema} if the Pravega metadata is needed.
+     *
+     * @param eventRead The EventRead structure the client returns which contains metadata
+     *
+     * @return the deserialized event with metadata
+     */
+    public T deserializeWithMetadata(EventRead<T> eventRead) {
```
claudiofahey: This name suggests that it replaces `deserialize` which is not the case. This method does not actually deserialize. I recommending renaming it to `insertMetadata` or something similar.

src/test/java/io/pravega/connectors/flink/FlinkPravegaReaderTest.java
```diff
@@ -433,6 +511,38 @@ public boolean isEndOfStream(Integer nextElement) {
         }
     }
 
+    /**
+     * A test JSON format deserialization schema with metadata.
+     */
+    private static class TestJsonDeserializationSchema extends PravegaDeserializationSchema<JsonNode> {
```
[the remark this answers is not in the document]
crazyzhou: The latest commits fix this. We use a POJO class `IntegerWithEventPointer` to hole the data and metadata.
It also exposes a general `JsonSerializer` in the package which I think might be useful to some 
users

src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
crazyzhou: It's true that this one does not do `deserialization`, but it really "get" the data. It has a default implementation `eventRead.getEvent()` and gets called in the `FlinkPravegaReader`  here: https://github.com/pravega/flink-connectors/pull/305/files#diff-005207bb16199a05351084fea79fd9acR271

crazyzhou: I'm also not a fan of the current name, but I don't think `insertMetadata` is correct because if it's not overriden which is most cases, it won't insert metadata. Maybe `getData` , `getEvent` is better? Any suggestions of a better name?

maddisondavid: When reviewing this code I also got confused by the name of this method because it's not actually obvious that both `deserialize` methods will get called, one at the Pravega level and one at the Flink Connector level.  

It's almost like the method should be called `extractEvent` since it's extracting it from the Pravega level API (and providing the chance to inject Metadata into the event).  I considered perhaps it should be called `getEvent` or `retrieveEvent` but this would get confusing with the Pravega semantics.

src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
```diff
@@ -29,6 +30,9 @@
  * <p>This adapter exposes the Pravega serializer as a Flink Deserialization schema and
  * exposes the produced type (TypeInformation) to allow Flink to configure its internal
  * serialization and persistence stack.
+ *
+ * <p>An additional method {@link #deserializeWithMetadata(EventRead)} is provided for
+ * applying the metadata in the deserialization. This method can be overriden in the extended class. </p>
  */
 public class PravegaDeserializationSchema<T>
```
maddisondavid: Type `T` here is overloaded in that it expresses both the type of object expected from the Pravega stream (used by the Pravega API) and the type of object expected from the connector (with possible Metadata injected).  This means that the object being stored in the stream MUST serialize to a Java class that expects that Metadata.  For some formats (such as JSON) this may be easy but for other formats such as Java it's difficult and messy.  In fact this can be seen in the Unit tests where `IntegerWithEventPointer` is the type being "stored" in the stream.

I'm not sure if it's possible, but it feels like there should be two types, a type `T` to represent the type from the stream and a type `C` to represent the type from the connector.  In most cases these could be the same but in other cases where MetaData needs to be injected these could be different.  

Or maybe (just thinking allowed) we're talking about two completely different interfaces, i.e. a PravegaMetaDataSchema<T, C> that extends PravegaDeserializationSchema<T> and just has the method 
```
public C deserializeWithMetadata(EventRead<T> eventRead) {
```

crazyzhou: Yes, this was my initial thought, and I also tried with it, but finally chose this. This is also what Kafka's approach to keep the same format of the event: https://github.com/apache/flink/blob/c6d9b37a8362317cc364c132ced5ac10dd3e10f2/flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/util/serialization/JSONKeyValueDeserializationSchema.java.

I wrote some investigation and the reason here: https://docs.google.com/document/d/1Dp8KYA6N81869dFjhD7w3z2hXd2QkU-vS1DNkz5aRC4/edit#

src/main/java/io/pravega/connectors/flink/serialization/PravegaDeserializationSchema.java
crazyzhou: `extractEvent` is a good one, I can buy in this name

PR #320, opened 2020-01-17
[issue-179] Deprecate `PravegaSerialization`

### LINKED ISSUE #179: Deprecate `PravegaSerialization`
**Problem description**
The `PravegaSerialization` is too non-specific about the serialization format that is used, and actually encourages the use of Java serialization which is very inefficient.   Let's leave the selection to the user.

**Problem location**

AlexanderZhao1:
**Change log description**
* `PravegaSerialization.java `

* `PravegaSerializationTest.java`

**Purpose of the change**
fix #179 

**What the code does**
As #179 mentioned,

> The PravegaSerialization is too non-specific about the serialization format that is used, and actually encourages the use of Java serialization which is very inefficient. Let's leave the selection to the user.

* Add  `Deprecated` annotation to PravegaSerialization and add deprecated java doc.
* Replace the use of PravegaSerialization in `PravegaSerializationTest` with JavaSerializer.

**How to verify it**

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/serialization/PravegaSerialization.java
+++ b/src/main/java/io/pravega/connectors/flink/serialization/PravegaSerialization.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2017 Dell Inc., or its subsidiaries. All Rights Reserved.
+ * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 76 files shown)


src/main/java/io/pravega/connectors/flink/serialization/PravegaSerialization.java
```diff
@@ -15,7 +15,12 @@
 /**
  * Helper methods to create DeserializationSchema and SerializationSchemas using the pravega JavaSerializer
  * for generic types.
+ *
+ * @deprecated PravegaSerialization is too non-specific about the serialization format that is used.
```
crazyzhou: In the javadocs, we should offer the alternative approach instead of explanation. 
```suggestion
 * @deprecated Please use the constructor with {@link JavaSerializer} instead.
```

PR #321, opened 2020-01-17
[issue-319] Handle TruncatedDataException in the reader

### LINKED ISSUE #319: TruncatedDataException on Job Restart
**Problem description**
When restarting a Job the Job will sometimes fail with a `TruncatedDataException`.  This exception is escaping into Flink and causing the Job to fail when the connector should really just capture the exception and retry the `readNextEvent` at which point Pravega will reset the offset to the correct point.

```
2020-01-16 12:20:22,999 INFO  org.apache.flink.runtime.executiongraph.ExecutionGraph        - Source: pravega-reader -> extract-number (1/5) (1791c4f5648bf46180408e4642034ac0) switched from RUNNING to FAILED.
io.pravega.client.stream.TruncatedDataException
        at io.pravega.client.stream.impl.EventStreamReaderImpl.handleSegmentTruncated(EventStreamReaderImpl.java:307)
        at io.pravega.client.stream.impl.EventStreamReaderImpl.readNextEventInternal(EventStreamReaderImpl.java:142)
        at io.pravega.client.stream.impl.EventStreamReaderImpl.readNextEvent(EventStreamReaderImpl.java:109)
        at io.pravega.connectors.flink.FlinkPravegaReader.run(FlinkPravegaReader.java:270)
        at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:94)
        at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:58)
        at org.apache.flink.streaming.runtime.tasks.SourceStreamTask.run(SourceStreamTask.java:99)
        at org.apache.flink.streaming.runtime.tasks.StoppableSourceStreamTask.run(StoppableSourceStreamTask.java:45)
        at org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:300)
        at org.apache.flink.runtime.taskmanager.Task.run(Task.java:704)
        at java.lang.Thread.run(Thread.java:748)
```

**Problem location**
https://github.com/pravega/flink-connectors/blob/master/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java#L270

Capture `TruncatedDataException` and retry the `readNextEvent`.  This exception should not escape into Flink.

crazyzhou:

**Change log description**
- Handle TruncatedDataException in the reader

**Purpose of the change**
Fixes #319 

**What the code does**
Let the reader call `readNextEvent` again when catching `TruncatedDataException`
Add tests to verify the exception is caught and read can go on as nornal

**How to verify it**
Target to master, should cherry-pick to all `0.7` and `0.6` branch

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -20,6 +20,7 @@ import io.pravega.client.stream.ReaderGroup;
 import io.pravega.client.stream.ReaderGroupConfig;
 import io.pravega.client.stream.Stream;
 import io.pravega.client.stream.StreamCut;
+import io.pravega.client.stream.TruncatedDataException;
 import io.pravega.connectors.flink.watermark.AssignerWithTimeWindows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.flink.api.common.ExecutionConfig;
@@ -267,7 +268,13 @@ public class FlinkPravegaReader<T>
 
             // main work loop, which this task is running
             while (this.running) {
-                final EventRead<T> eventRead = pravegaReader.readNextEvent(eventReadTimeout.toMilliseconds());
+                EventRead<T> eventRead;
+                try {
+                    eventRead = pravegaReader.readNextEvent(eventReadTimeout.toMilliseconds());
+                } catch (TruncatedDataException e) {
+                    // Data is truncated, Force the reader going forward to the next available event
+                    continue;
+                }
                 final T event = eventRead.getEvent();
 
                 // emit the event, if one was carried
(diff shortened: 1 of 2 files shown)


src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -267,7 +268,13 @@ public void run(SourceContext<T> ctx) throws Exception {
 
             // main work loop, which this task is running
             while (this.running) {
-                final EventRead<T> eventRead = pravegaReader.readNextEvent(eventReadTimeout.toMilliseconds());
+                EventRead<T> eventRead;
+                try {
+                    eventRead = pravegaReader.readNextEvent(eventReadTimeout.toMilliseconds());
+                } catch (TruncatedDataException e) {
+                    // Data is truncated, Force the reader going forward to the next available event
+                    continue;
```
maddisondavid: Hopefully this never gets into a situation where it loops forever!

PR #344, opened 2020-03-12
[Issue-343] Inclusive vs. Exclusive Watermarks

### LINKED ISSUE #343: Inclusive vs. Exclusive Watermarks
**Problem description**
The code of `LowerBoundAssigner` subtracts one from the lower bound to determine the watermark:

https://github.com/pravega/flink-connectors/blob/master/src/main/java/io/pravega/connectors/flink/watermark/LowerBoundAssigner.java#L25

It sounds like that was done this way so that the watermark is inclusive, inclusive in the sense that the low watermark _t_ indicates that all events with timestamps smaller or equal to _t_ have been read. That's is confusing, though, and it does not seem aligned with how Flink deals with watermarks.

To explain the issue, let me describe the problem I came across. In my experiment, I created 50 files with 1,000 events in each file. To simplify the time assignment, I have assigned the same timestamp to all events in a file, and the timestamp is the file number. The watermark here is supposed to represent the progression of files. I have some simple window aggregation and each window corresponds to processing the events with a single timestamp (same file).

If I note time at the end of each file with the file number representing time, then after processing all files, I only get 49 windows processed, and that's because the watermark never progresses beyond the value that closes the last window because we are subtracting -1L.

I did a second experiment to validate that I get all windows processed when I use file number + 1. In that case, I eventually get a watermark that is equal to the number of the last file and I get all expected windows processed.

According to my evaluation, Flink treats the watermark in an inclusive manner, like I defined above. If so, we should remove the -1 subtraction to comply with that expectation.

**Problem location**
`LowerBoundAssigner`

Remove the subtraction.

Guangfeng-Xu:
**Change log description**

- Change the code of `LowerBoundAssigner` so that it doesn't subtract one from the lower bound to determine the watermark
- Change the test code corresponding to `LowerBoundAssigner`

**Purpose of the change**
Fix #343 

**What the code does**
Flink treats the watermark in an inclusive manner, so the -1 subtraction should be removed to comply with that expectation.

**How to verify it**

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/watermark/LowerBoundAssigner.java
+++ b/src/main/java/io/pravega/connectors/flink/watermark/LowerBoundAssigner.java
@@ -31,7 +31,7 @@ public abstract class LowerBoundAssigner<T> implements AssignerWithTimeWindows<T
         }
         return timeWindow.getLowerTimeBound() == Long.MIN_VALUE ?
                 new Watermark(Long.MIN_VALUE) :
-                new Watermark(timeWindow.getLowerTimeBound() - 1L);
+                new Watermark(timeWindow.getLowerTimeBound());
     }
 }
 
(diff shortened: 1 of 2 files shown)


fpj:
The change looks good, but one question I have is with respect to documentation. Are we documenting anywhere how to use our watermarking support with Flink jobs? I ask because the important change here is the semantic one, not the code change itself, and we would need to reflect this semantic change somewhere.

crazyzhou:
Yes, we have the documentation here: 
https://github.com/pravega/flink-connectors/blob/master/documentation/src/docs/streaming.md#timestamp-extraction-watermark-emission

https://github.com/pravega/flink-connectors/blob/master/documentation/src/docs/streaming.md#watermark

PR #349, opened 2020-03-17
[Issue-69] Replace Guava preconditions with Flink preconditions

### LINKED ISSUE #69: Prefer Flink annotations and preconditions
**Problem description**
With the rationale that the connector code should sit nicely with that of other Flink connectors, prefer Flink annotations over Guava annotations where possible.  Concretely: `Preconditions`, `VisibleForTesting`.

Note that the Pravega client itself uses Guava, so it will remain a dependency.

AlexanderZhao1:
**Change log description**
* change all `com.google.common.base.Preconditions` to `org.apache.flink.util.Preconditions`

**Purpose of the change**
Fix #69 

**What the code does**
Replace Guava Preconditions with Flink Preconditions

**How to verify it**
`./gradlew clean build`should pass


src/main/java/io/pravega/connectors/flink/ReaderCheckpointHook.java
```diff
@@ -24,8 +24,9 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
+import static org.apache.flink.util.Preconditions.checkArgument;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
```
crazyzhou, line 29: Remove this extra line

PR #365, opened 2020-04-17
[Issue 364] Make `writer.enable-watermark` optional in Table API properties

### LINKED ISSUE #364: Make `writer.enable-watermark` optional in Table API properties
**Problem description**
In the current code of Table API,  `writer.enable-watermark` is required. It means that when users are using Flink SQL client with a YAML environmental file, or create a sink with properties map, the `writer.enabled-watermark` section is required explicitly.
As the best practice similar with other connectors https://ci.apache.org/projects/flink/flink-docs-release-1.10/dev/table/connect.html , we should put this optional feature as an optional property in the map. 
This also makes our current documentation of the Table API [here](https://github.com/pravega/flink-connectors/blob/master/documentation/src/docs/table-api.md#sample-environment-file) correct without getting exception of missing the required property: `writer.enable-watermark`. The document upgrade will be addressed in #357 

**Problem location**
Table API

Should change it to `Optional<Boolean>`.

Guangfeng-Xu:
Signed-off-by: Guangfeng Xu Guangfeng.Xu@dell.com

**Change log description**

1. Change `watermark` in `ConnectorConfigurations.java` from type `Boolean` to type `Optional<Boolean>`
2. Use the `ConnectorConfigurations` to decide whether to enable `watermark` or not

**Purpose of the change**
Fix #364 

**What the code does**
Change the default required `writer.enable-watermark` to optional

**How to verify it**


crazyzhou:
@fpj It looks like the coverage is decreased because we removed a covered line and added an `if` on it. The `if` line is dragging the percentage. Can you please have a review and check if we can let it pass this time.

PR #375, opened 2020-05-19
[Issue-374] withDefaultScope bugfix

### LINKED ISSUE #374: PravegaConfig.withDefaultScope should always set the default scope
**Problem description**
`PravegaConfig.withDefaultScope` currently ignores the provided value if the scope has already been set, possibly by the environment variable PRAVEGA_SCOPE. However, this makes it impossible to change the default scope if PRAVEGA_SCOPE has been defined. The Pravega default scope is not _just_ a default value that can be overridden when creating a stream. It is also used as the scope in which all reader groups will be created.

The motivation for this change is as follows. Consider a Flink job that runs in Kubernetes with PRAVEGA_SCOPE=scope1 but which needs to read from a stream in scope2 on a different Pravega controller. When it is tries to create the reader group, it will use the controller for scope2 but the "default" scope scope1. This will fail.

This change will also eliminate the inconsistent behavior of this method compared to similar `with...` methods in the class.

**Problem location**
PravegaConfig.withDefaultScope

Remove check:
```
if (this.defaultScope == null)
```

AlexanderZhao1:
Signed-off-by: AlexanderZhao1 <Alexander_Zhao@dell.com>

**Change log description**

`PravegaConfig.withDefaultScope`

`PravegaConfigTest.java `

**Purpose of the change**

Fixes #374 

**What the code does**

remove `if (this.defaultScope == null)` to make sure override the default scope is possible.

**How to verify it**

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/PravegaConfig.java
+++ b/src/main/java/io/pravega/connectors/flink/PravegaConfig.java
@@ -137,6 +137,18 @@ public class PravegaConfig implements Serializable {
      * @param scope The scope to use (with lowest priority).
      */
     public PravegaConfig withDefaultScope(String scope) {
+        if (this.defaultScope == null){
+            this.defaultScope = scope;
+        }
+        return this;
+    }
+
+    /**
+     * Configures the self-defined Pravega scope
+     *
+     * @param scope The scope to use.
+     */
+    public PravegaConfig withScope(String scope) {
         this.defaultScope = scope;
         return this;
     }
--- a/src/main/java/io/pravega/connectors/flink/PravegaConfig.java
+++ b/src/main/java/io/pravega/connectors/flink/PravegaConfig.java
@@ -137,9 +137,7 @@ public class PravegaConfig implements Serializable {
      * @param scope The scope to use (with lowest priority).
      */
     public PravegaConfig withDefaultScope(String scope) {
-        if (this.defaultScope == null) {
-            this.defaultScope = scope;
-        }
+        this.defaultScope = scope;
         return this;
     }
 
--- a/src/main/java/io/pravega/connectors/flink/PravegaConfig.java
+++ b/src/main/java/io/pravega/connectors/flink/PravegaConfig.java
@@ -146,7 +146,7 @@ public class PravegaConfig implements Serializable {
     /**
      * Configures the self-defined Pravega scope.
      *
-     * @param scope The scope to use.
+     * @param scope The scope to use (with highest priority).
      */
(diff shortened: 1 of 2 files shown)


src/main/java/io/pravega/connectors/flink/PravegaConfig.java
```diff
@@ -137,9 +137,7 @@ public PravegaConfig withTrustStore(String trustStore) {
      * @param scope The scope to use (with lowest priority).
      */
     public PravegaConfig withDefaultScope(String scope) {
```
crazyzhou, line 139: I prefer change this into `withScope` and the variable into `scope`. But this would be breaking change, is there any concern on this? @maddisondavid

src/main/java/io/pravega/connectors/flink/PravegaConfig.java
```diff
@@ -137,9 +137,7 @@ public PravegaConfig withTrustStore(String trustStore) {
      * @param scope The scope to use (with lowest priority).
```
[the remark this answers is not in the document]
crazyzhou: We agreed to fix this with a new `withScope` API

src/main/java/io/pravega/connectors/flink/PravegaConfig.java
crazyzhou: To keep compatibility, we keep the original and add a new `withScope` to do this.

src/test/java/io/pravega/connectors/flink/PravegaConfigTest.java
```diff
@@ -78,6 +78,19 @@ public void testScopePriority() {
         assertEquals("scope2", config.getDefaultScope());
     }
 
+    @Test
+    public void testScopeSet() {
```
crazyzhou, line 82: We can add this case into the above one. This is also part of the priority check.

src/main/java/io/pravega/connectors/flink/PravegaConfig.java
```diff
@@ -143,6 +143,16 @@ public PravegaConfig withDefaultScope(String scope) {
         return this;
     }
 
+    /**
+     * Configures the self-defined Pravega scope.
+     *
+     * @param scope The scope to use.
```
crazyzhou, line 149: Add that it is in the highest priority

crazyzhou:
Note that this is a change which will cherry-pick to `0.7` and `0.8` branches, so 0.7.1 release will include it. We keep the `withDefaultScope` method to keep compatibility. We will reconsider in the next release cycle if we really need this.

PR #381, opened 2020-06-04
[issue-377] Refactor Pravega Sink to extend TwoPhaseCommitSinkFunction

### LINKED ISSUE #377: Refactor transactional writer to use Flink TwoPhaseCommitSink interface
**Problem description**
`FlinkPravegaWriter` has similar code with Flink `TwoPhaseCommitSink` interface. Flink suggests
> This is a recommended base class for all of the SinkFunction that intend to implement exactly-once semantic.

and the interface `ListCheckpointed` the writer use now is marked deprecated in the coming 1.11 release. We should do this refactoring.

**Problem location**
Transactional writers in `FlinkPravegaWriter`.

Try to keep the backward compatibility.

crazyzhou:

**Change log description**
Refactor Pravega Sink to extend `TwoPhaseCommitSinkFunction`

**Purpose of the change**
Fixes #377 

**What the code does**
Refactor Pravega Sink to extend `TwoPhaseCommitSinkFunction.`
Keep the Pravega writers as a member of `FlinkPravegaWriter` function.
Keep Pravega transaction, transactionId and watermark in the state for recovery.
Have no idea why this supported recovery with transactions on diferent multiple streams. remove the support here.

**How to verify it**


crazyzhou:
@fpj I tried my best to add both positive and negative test cases to increase the coverage, and we can see the codecov for patch has already reach the requirement. But the project coverage is still decreasing because of the removed covered lines. Can you please add this to an exception?

maddisondavid:
While the code looks OK, would the change in the saved checkpoint format cause an upgrade issue for anybody attempting to restore from a savepoint created with a previous version connector?  i.e. Would this cause an upgrade breaking change?

crazyzhou:
Yes, as the transaction state is changed from the `PendingTransactions` into the new `PravegaTransactionState`, this would cause an breaking change for upgrade(savepoint recovery). My suggestion is to only include this change to the new 1.11 connector.

maddisondavid:
My understanding is that we expect the upgrade to Flink 1.11 connector to include some API breaking changes, therefore it SEEMS reasonable that this breaking change could also be included.  It would be good to get @fpj input into whether he would consider this an acceptable compromise.

crazyzhou:
@maddisondavid Tested the upgrade from 1.10 into 1.11 with new connector, the old savepoints can be compatible with the new application with the new connector of this PR. See the below Flink UI which tells that it recovers from last externalized checkpoint `chk-3`, and continue ingestion.
image

maddisondavid (approved):
LGTM

PR #406, opened 2020-09-04
[Issue-405] Call Close on Pravega Client Factory

### LINKED ISSUE #405: Pravega ClientFactory is not Closed when Reader closes
**Problem description**
During construction of a Reader a Pravega `ClientFactory` is implicitly created as part of creating a `ReaderGroupManager` or an `EventStreamClientFactory`.  Creation of the `ClientFactory` creates a Thread and Connection Pool that are only shutdown when `close` is explictly called.

Unfortunately in some places the reference to the `ReaderGroupManager` or `EventStreamClientFactory` isn't being stored and therefore `close` is not called when the reader is closed.  This results in these classes being left running even after the Reader task has been cancelled by Flink.  The outcome of this is that the Flink ChildFirstClassLoader cannot be garbage collected eventually resulting in the JVM Metaspace being exhausted if an application deployment is retried multiple times.

**Problem location**
Two places are an issue:
- *EventStreamClientFactory* https://github.com/pravega/flink-connectors/blob/master/src/main/java/io/pravega/connectors/flink/util/FlinkPravegaUtils.java#L106-L107
- *ReaderGroupManager* https://github.com/pravega/flink-connectors/blob/master/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java#L541-L556

When creating either the `ReaderGroupManager` or the `EventStreamClientFactory` a reference to the objects should be stored and `close` should be called on the objects during `FlinkPravegaReader::close`

maddisondavid:
Signed-off-by: David Maddison <david.maddison@dell.com>

**Change log description**
- Call Close on EventStreamClientFactory
- Call Close on ReaderGroupManager
- Update Unit Tests

**Purpose of the change**
Fix #405

**What the code does**
Storing references to created EventStreamClientFactory and ReaderGroupManager and calls `close` on them when the reader is closed.  This allows the Flink ClassLoaders to be successfully garbage collected removing the `OutOfMemory: Metaspace` error that can sometimes be hit if a Flink Job deployment is retried multiple times.

**How to verify it**
`FlinkPravegaReaderTest` Unit test was updated and successfully passes.

While it doesn't itself fix the Unit test issues, this is an important step forward in closing down any Pravega Client ThreadPools


shrids:
Should'nt  `this.batchClientFactory ` be closed too. ref: https://github.com/pravega/flink-connectors/blob/deca25ac3bf161bbaf6142dee9e3347cfe7027f2/src/main/java/io/pravega/connectors/flink/FlinkPravegaInputFormat.java#L171-L173

Given that BatchClientFactory spawns a connection pool which spawns the clientInternal threads.

maddisondavid:
Possibly, BUT our tests (i.e. Flink Longevity) don't use any of the Table or Batch APIs, it's all streaming readers

crazyzhou (approved):
LGTM. Approve it for now to unblock the test, we will raise another PR to focus on close calls on Table and batch API.

PR #410, opened 2020-09-16
[issue-409] Fix Dynamic-loading credentials for Pravega client in Table API

### LINKED ISSUE #409: Dynamic-loading credentials for Pravega client cannot work in Table API
**Problem description**
In Table API, we simply treat all the credentials as static ones as this. https://github.com/pravega/flink-connectors/blob/77dc893c1ce2fac4b597843424e1b3ae592e29b5/src/main/java/io/pravega/connectors/flink/util/ConnectorConfigurations.java#L227 
This method will record the token as a static string, which raises the authentication problem when we use dynamic-loading credentials like keycloak after token refresh.

**Problem location**
Table API
`ConnectorConfigurations::createPravegaConfig`

- Check for the value of the env variable `pravega.client.auth.loadDynamic` and system property `pravega_client_auth_loadDynamic`.
  - If at least one of them is `true`, the intent of the context is to load things dynamically from the environment or system properties, so we can skip the `SimpleCredentials()` piece and let the ClientConfig logic figure itself out.
  - If neither of them is present or `true`, it's safe to assume someone set them explicitly and they're very likely to be some kind of credentials that don't need refreshing (like Basic), and passing them along using the `SimpleCredentials()` seems like the right thing to do.

crazyzhou:
**Change log description**
Add check for Dynamic-loading credentials for Pravega in Table API

**Purpose of the change**
Fixes #409 

**What the code does**
Check for the value of the env variable `pravega.client.auth.loadDynamic` and system property `pravega_client_auth_loadDynamic` to decide if we keep the credential as `SimpleCredentials`.

**How to verify it**

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/util/ConnectorConfigurations.java
+++ b/src/main/java/io/pravega/connectors/flink/util/ConnectorConfigurations.java
@@ -63,6 +63,8 @@ import static io.pravega.connectors.flink.table.descriptors.Pravega.CONNECTOR_WR
  */
 @Data
 public final class ConnectorConfigurations {
+    private static final String AUTH_PARAM_LOAD_DYNAMIC = "pravega.client.auth.loadDynamic";
+    private static final String AUTH_PARAM_LOAD_DYNAMIC_ENV = "pravega_client_auth_loadDynamic";
 
     private Optional<Boolean> metrics;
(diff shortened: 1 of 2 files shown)


maddisondavid (approved):
LGTM

PR #414, opened 2020-09-30
[issue-413] Ensure the reader name below the length limit

### LINKED ISSUE #413: Flink Reader ID will exceed length limit
**Problem description**
Pravega client 0.8 added a length check of 256 of the reader ID. But in SQL use cases, there will be long SQL queries to be the task name for the reader, which may violate this check.

**Problem location**
`FlinkPravegaUtils`

Add a check and truncate for too-long task name to generate reader

youminhan:
**Change log description**
Add a check and truncate for too-long task name to generate reader ID.

**Purpose of the change**
Fixes #413 

**What the code does**
Pravega client 0.8 added a length check of 256 of the reader ID. But in SQL use cases, there will be long SQL queries to be the task name for the reader, which may violate this check. Add a check and truncate for too-long task name to generate reader ID.

**How to verify it**
`./gradlew clean build ` passes

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/util/FlinkPravegaUtils.java
+++ b/src/main/java/io/pravega/connectors/flink/util/FlinkPravegaUtils.java
@@ -62,7 +62,11 @@ public class FlinkPravegaUtils {
      * @return the generated default reader name.
      */
     public static String getReaderName(final String taskName, final int index, final int total) {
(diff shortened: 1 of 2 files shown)


crazyzhou (approved):
LGTM. Thanks for the contribution and it's great to fix it just before 0.8 release.

PR #421, opened 2020-09-30
[issue-416] Retry close when InterruptedException is caught in Pravega reader

### LINKED ISSUE #416: Zombie GRPC threads in job cancel
**Problem description**
There are constant `grpc-default-worker-ELG` threads left runnable each time we cancel the Flink job with Flink Pravega reader. These threads hold the classloader GC, so that there will be metaspace OOM error during continuous job restarts.

_**Root cause:**_
Flink calls the interruption in the TaskCanceler https://github.com/apache/flink/blob/43f82614fa66a2d25edf51297147d5b341f9e076/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java#L1095.
This brings a race condition with the connector's `close` method. If the `close` method on the connector can complete it's work BEFORE the TaskCanceler calls Interrupted, then we get a clean disposal with no zombie threads. However, the chance of this is quite low which is why we will constantly have zombie threads with 

**Problem location**
`FlinkPravegaReader::close`

Retry close if InterruptedException is caught.

crazyzhou:
**Change log description**
- Update Pravega to use submodule to pick up the fix needed (https://github.com/pravega/pravega/pull/5226)
- Retry close when InterruptedException is caught in Pravega reader
- Record the last exception in close method and throw it 

**Purpose of the change**
Fixes #416 

**What the code does**
A "cleaner" close for `FlinkPravegaReader`

**How to verify it**
Tested locally with Flink and Pravega standalone cluster with reader application, no more GRPC threads after the job cancel.
Target to `r0.8` now, should cherry-pick to all `0.8` branches.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -360,19 +360,48 @@ public class FlinkPravegaReader<T>
 
     @Override
     public void close() throws Exception {
+        Throwable lastThrowable = null;
         if (eventStreamClientFactory != null) {
-            log.info("Closing Pravega eventStreamClientFactory");
-            eventStreamClientFactory.close();
+            try {
+                log.info("Closing Pravega eventStreamClientFactory");
+                eventStreamClientFactory.close();
+            } catch (Throwable e) {
+                if (e instanceof InterruptedException) {
+                    log.warn("Interrupted while waiting for eventStreamClientFactory to close, retrying ...");
+                    eventStreamClientFactory.close();
+                } else {
+                    lastThrowable = e;
+                }
+            }
         }
-
         if (readerGroupManager != null) {
             log.info("Closing Pravega ReaderGroupManager");
-            readerGroupManager.close();
+            try {
+                readerGroupManager.close();
+            } catch (Throwable e) {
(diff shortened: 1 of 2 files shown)


maddisondavid (approved):
LGTM

src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -360,19 +361,48 @@ public void open(Configuration parameters) throws Exception {
...
-            eventStreamClientFactory.close();
+            try {
+                log.info("Closing Pravega eventStreamClientFactory");
+                eventStreamClientFactory.close();
+            } catch (Throwable e) {
+                if (e instanceof InterruptedException) {
+                    log.warn("Interrupted while waiting for eventStreamClientFactory to close, retrying ...");
+                    eventStreamClientFactory.close();
+                } else {
+                    ex = ExceptionUtils.firstOrSuppressed(e, ex);
```
crazyzhou, line 374: I changed to use Flink suggested way to expose exceptions in close method.
ref: https://github.com/apache/flink/blob/4467277dcf80d18b2898d4ec93c634dc33d25b2d/flink-core/src/main/java/org/apache/flink/util/ExceptionUtils.java#L253

PR #444, opened 2021-01-29
[issue-443] Upgrade to Flink 1.10.3

### LINKED ISSUE #443: GenericRecord deserialization test case fails on Flink 1.10 connector
**Problem description**
On cherry-picking the commits into the Flink 1.10 connector, the `testInputFormatWithAvroRegistryDeserializer` is failing with below error
```
The class 'org.apache.avro.generic.GenericRecord' is not instantiable: The class is not a proper class. It is either abstract, an interface, or a primitive type.
```
It is the same issue described in https://issues.apache.org/jira/browse/FLINK-18223 and fixed in `1.10.2`, but we are still using `1.10.0`. We need to bump up the Flink version of Flink 1.10 connector to fix this

**Problem location**
`gradle.properties`

crazyzhou:
**Change log description**
- Upgrade to Flink 1.10.3

**Purpose of the change**
Fixes #443 

**What the code does**
Upgrade to Flink 1.10.3

**How to verify it**

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@
 
 # 3rd party Versions.
 checkstyleToolVersion=7.1
-flinkVersion=1.10.0
+flinkVersion=1.10.3
 flinkScalaVersion=2.12
 jacksonVersion=2.8.9
 lombokVersion=1.18.4


maddisondavid (approved):
LGTM

PR #453, opened 2021-02-16
Change FlinkPravegaDynamicTableITCase to use a single segment

### LINKED ISSUE #452: Fix faulty test FlinkPravegaDynamicTableITCase
**Problem description**
This test writes events to Pravega and then runs a Flink batch job to read the events. It assumes that it will read all rows immediately but Pravega makes no such guarantee.

**Problem location**
FlinkPravegaDynamicTableITCase.java

Run Job_2 (read job) repeatedly until the result matches the expected value.

claudiofahey:
**Change log description**
Change FlinkPravegaDynamicTableITCase to use a single segment.

**Purpose of the change**
This integration test appears to be using watermarks in a way that is dependent on the order in which events are written across routing keys. Since the order across routing keys is not guaranteed by Pravega, this causes the test to fail intermittently. 
Fixes #452 

**What the code does**
The change in this PR changes the number of segments from 3 to 1, avoiding ordering issues. Proper ordering and watermarking should be tested as part of a different integration test.

**How to verify it**
```
./gradlew build
```

### DIFF
--- a/src/test/java/io/pravega/connectors/flink/FlinkPravegaWriterITCase.java
+++ b/src/test/java/io/pravega/connectors/flink/FlinkPravegaWriterITCase.java
@@ -180,7 +180,6 @@ public class FlinkPravegaWriterITCase {
                 }
                 actualEventCount++;
             }
-            Assert.assertEquals(actualEventCount, EVENT_COUNT_PER_SOURCE);
             if (EVENT_COUNT_PER_SOURCE == actualEventCount) {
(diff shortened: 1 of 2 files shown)


pavanbhat21 (approved):
LGTM

jadnhm (approved):

shrids (approved):
LGTM.
(Merging it since Brian and David are on vacation)

PR #476, opened 2021-03-22
[Issue-475] Fail to write the log in the tests

### LINKED ISSUE #475: Fail to write the log in the tests
**Problem description**

The project is configured to write the log via logback during the tests, however, it doesn't work for now. We need to fix this.

**Problem location**

`build.gradle` and `java.gradle`.

Use the existing `log4j2` module.

thekingofcity:

**Change log description**

Use `log4j2` to write logs.

**Purpose of the change**

Fixes #475 

**What the code does**

* Remove the useless legacy log settings `testLogging`.
* Add `log4j2` properties.

---

Most tutorials[1] [2] will assume a `log4j2.properties` file under the classpath will work, however, only when the file is named `log4j2-test.properties` [3] will work. (Because the type is test?)
For tutorials [4] [5] involve how to set the system property, they use `systemProperty "key", "value"` which doesn't work for unknown reason. The alternative `systemProperties "key": "value"` doesn't work either.

So an usable solution is to use the above `System.properties["key"] = "value"` syntax.
`System.properties["log4j2.properties"] = new File(buildDir,'resources/test/log4j2-test.xml').absolutePath`

Or simply name the file `log4j2-test.properties`. **Selected Solution**

PS: `println findProperty("log4j2.configurationFile")` will still complain null, which is weird.

[1]: https://springframework.guru/log4j-2-configuration-using-properties-file/
[2]: https://tomgregory.com/gradle-project-properties-best-practices/
[3]: https://www.scalyr.com/blog/log4j2-configuration-detailed-guide/
[4]: https://stackoverflow.com/questions/17851413/gradle-doesnt-pass-system-properties-to-test-classes
[5]: https://stackoverflow.com/questions/21406265/how-to-give-system-property-to-my-test-via-gradle-and-d

**How to verify it**

Passes `./gradlew clean build` and has `connector-test.log` in the root directory after testes.


src/test/resources/log4j2-test.properties
```diff
@@ -0,0 +1,41 @@
+# Allows this configuration to be modified at runtime. The file will be checked every 30 seconds.
```
crazyzhou, line 1: With this file, we can remove the logback configurations in this folder now.

crazyzhou (changes requested):

src/test/resources/log4j2-test.properties
thekingofcity: Removed.

crazyzhou (approved):
LGTM

PR #482, opened 2021-04-01
[Issue-383] Change EventTimeOrderingOperator into EventTimeOrderingFunction

### LINKED ISSUE #383: Change `EventTimeOrderingOperator` into `EventTimeOrderingFunction`
**Problem description**
`EventTimeOrderingOperator` extends the `OneInputStreamOperator`, which is non-keyed while keyed state is used in the implementation. It is not ideal and even not allowed in Flink 1.11 due to https://issues.apache.org/jira/browse/FLINK-16316 . 

**Problem location**
`EventTimeOrderingOperator`

Change the class to `EventTimeOrderingFunction` which extends `KeyedProcessFunction`, as this allows to use keyed state. We should also change the implementation inside the `FlinkPravegaUtils.writeToPravegaInEventTimeOrder` function and keep the compatibility.

thekingofcity:

**Change log description**

EventTimeOrderingOperator extends the OneInputStreamOperator, which is non-keyed while keyed state is used in the implementation. It is not ideal and even not allowed in Flink 1.11 due to https://issues.apache.org/jira/browse/FLINK-16316 .

**Purpose of the change**

Fixes #383 

**What the code does**

* Change the class to `EventTimeOrderingFunction` which extends `KeyedProcessFunction`, as this allows to use keyed state.
* Change the implementation inside the `FlinkPravegaUtils.writeToPravegaInEventTimeOrder` function and keep the compatibility.
* Update the corresponding tests.
* Use `ProcessFunctionTestHarnesses.forKeyedProcessFunction` to create the harness for the function. [Ref](https://ci.apache.org/projects/flink/flink-docs-stable/dev/stream/testing.html#unit-testing-processfunction)

**How to verify it**


src/main/java/io/pravega/connectors/flink/EventTimeOrderingFunction.java
```diff
@@ -0,0 +1,125 @@
...
+    public EventTimeOrderingFunction(TypeInformation<T> typeInformation) {
+        this.typeInformation = typeInformation;
+    }
+
+    @Override
+    public void open(Configuration config) throws Exception {
+        super.open(config);
+
+        // create a map-based queue to buffer input elements
+        if (elementQueueState == null) {
```
crazyzhou, line 58: It is redundant to have the null check here, we can always get the state from the context.

src/main/java/io/pravega/connectors/flink/EventTimeOrderingFunction.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    @Override
+    public void processElement(T element, Context ctx, Collector<T> out) throws Exception {
+        // timestamp of the processed element
+        Long timestamp = ctx.timestamp();
+        if (timestamp == null) {
+            // elements with no time component are simply forwarded.
+            // likely cause: the time characteristic of the program is not event-time.
```
crazyzhou, line 81: There are processing time, ingestion time and event time here, and only processing time is making the `timestamp` null here, so you can change the comment as 
```
Simply forward the elements when the time characteristic of the program is set to ProcessingTime
```

src/main/java/io/pravega/connectors/flink/EventTimeOrderingFunction.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * The input type information for buffering events to managed state.
+     */
+    private final TypeInformation<T> typeInformation;
+
+    /**
+     * The queue of input elements keyed by event timestamp.
+     */
+    private transient MapState<Long, List<T>> elementQueueState;
```
crazyzhou, line 42: It is a bit strange to see a map state with a comment saying it's a queue. It is actually a map contains all the data between watermarks. We can use the word dataState along with the name there.

src/test/java/io/pravega/connectors/flink/EventTimeOrderingFunctionTest.java
```diff
@@ -104,15 +100,15 @@ public void testOrdering() throws Exception {
     @Test
     public void testLateElements() throws Exception {
```
crazyzhou, line 101: This test does not really test late events in such case because there are no data triggering the timestamp moving forward before the watermark. We need to change this test a little bit.
First we can let an on-time data go in the function, and then comes the watermark, check the output, and then process another late data to see if it is discarded.

crazyzhou (changes requested):

src/main/java/io/pravega/connectors/flink/EventTimeOrderingFunction.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.typeutils.ListTypeInfo;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.streaming.api.functions.KeyedProcessFunction;
+import org.apache.flink.util.Collector;
+
+import java.util.ArrayList;
+import java.util.List;
+
```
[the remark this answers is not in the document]
thekingofcity: Added

src/main/java/io/pravega/connectors/flink/EventTimeOrderingFunction.java
thekingofcity: remove the check and create the map whenever the `open` is called

src/main/java/io/pravega/connectors/flink/EventTimeOrderingFunction.java
thekingofcity: Changed

src/main/java/io/pravega/connectors/flink/EventTimeOrderingFunction.java
thekingofcity: Change the name to `dataState` with comment `State to buffer all the data between watermarks.`

src/test/java/io/pravega/connectors/flink/EventTimeOrderingFunctionTest.java
thekingofcity: Updated.

crazyzhou (approved):
LGTM

PR #483, opened 2021-04-02
[Issue-481] Row-based router cannot be specified on non-null field

### LINKED ISSUE #481: Row-based router cannot be specified on non-null field
**Problem description**
In the new table sink, the non null fields cannot be specified as a routing key filed.
It is caused by the check 
```
checkArgument(DataTypes.STRING().equals(fieldTypes[keyIndex]), "Key field must be of type 'STRING'");
```
Comparing on `DataType` level is not correct in this case, we need to check if the logical type root is `CHAR` or `VARCHAR` to determine if it is a string or not.

**Problem location**
`FlinkPravegaDynamicTableSink`

thekingofcity:

**Change log description**

In the new table sink, the non null fields cannot be specified as a routing key filed.
It is caused by the check

```java
checkArgument(DataTypes.STRING().equals(fieldTypes[keyIndex]), "Key field must be of type 'STRING'");
```

Comparing on DataType level is not correct in this case, we need to check if the logical type root is `CHAR` or `VARCHAR` to determine if it is a string or not.

**Purpose of the change**

Fixes #481 

**What the code does**

Check the type's root so both string and other string-like type like `CHAR` and `VARCHAR` could pass the check.

**How to verify it**

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/dynamic/table/FlinkPravegaDynamicTableSink.java
+++ b/src/main/java/io/pravega/connectors/flink/dynamic/table/FlinkPravegaDynamicTableSink.java
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
  */
 package io.pravega.connectors.flink.dynamic.table;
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/pravega/connectors/flink/dynamic/table/FlinkPravegaDynamicTableSink.java
```diff
@@ -1,11 +1,11 @@
 /**
  * Copyright (c) Dell Inc., or its subsidiaries. All Rights Reserved.
- *
+ * <p>
```
crazyzhou, line 3: Is there a reason you added `<p>` here? It seems we don't have this tag in other files.

thekingofcity: That's strange. I didn't change it during developing

thekingofcity:
Remove the `<p>` in `EventTimeOrderingFunction`

crazyzhou (approved):
LGTM

PR #489, opened 2021-04-13
[Issue-145] Add pravega client dependency to connector POM

### LINKED ISSUE #145: Add pravega client dependency to connector POM
**Problem description**
The Pravega client is currently shaded but not relocated in the connector JAR, and is not mentioned in the effective POM.  This is an odd situation, because an application that links to the connector is not told that Pravega is included, yet it is, or what version is included.   The application may well need to use the client directly, for example the pravega samples pre-create a scope in their main method.

Adding the client dependency is not as simple as it sounds, however.    If the connector were to have a straightforward dependency on the Pravega client in its effective POM, the application would see it as a transitive dependency and include it in its own uberjar _without the shading that is normally required_.    The true prerequisite is pravega/pravega#1339 (Pravega client to shade its dependencies), which would vastly simplify both the connector and the application.

**Problem location**

- fix pravega/pravega#1339
- eliminate shading logic in connector 
- declare pravega client dependency

thekingofcity:

**Change log description**

Remove the manual pom dependencies injections so that the created pom will have the right dependencies.

**Purpose of the change**

Fixes #145 

**What the code does**

* Remove the manual pom dependencies injections.
* Also remove the `bintray.gradle` that nobody uses.

**How to verify it**

`gradle install` will generate pom with right dependencies mentioned in the `build.gradle`.
This should also affect the `publishToRepo` task.


crazyzhou (approved):
LGTM

PR #495, opened 2021-04-19
[Issue-493] Make reader group optional in table API

### LINKED ISSUE #493: Make reader group optional in table API
**Problem description**
Reader group name is now a required option in the new table API. With some more users trying that, we realized that such constraint disallows the table to be reused for read, otherwise it will have `ReaderAlreadyInReaderGroup` Exception. We need to change the reader group name to optional just like we do for normal DataStream source.

**Problem location**
Table API

thekingofcity:

**Change log description**

Reader group name is now a required option in the new table API. With some more users trying that, we realized that such constraint disallows the table to be reused for read, otherwise it will have ReaderAlreadyInReaderGroup Exception. We need to change the reader group name to optional just like we do for normal DataStream source.

**Purpose of the change**

Fixes #493 

**What the code does**

* Remove the null check during the initialization.
* Make the reader group name optional by setting it if it is null.
* Test the auto-generated random group name in `AbstractStreamingReaderBuilder`.
* Test the initialization function call with group name parameter setting to null.

**How to verify it**


crazyzhou (approved):
LGTM

src/main/java/io/pravega/connectors/flink/dynamic/table/FlinkPravegaDynamicTableSource.java
```diff
@@ -114,6 +112,31 @@ public FlinkPravegaDynamicTableSource(DataType producedDataType,
         this.isBounded = isBounded;
     }
 
+    public FlinkPravegaDynamicTableSource(DataType producedDataType,
```
crazyzhou, line 115: This part is redundant, we can remove such constructor

thekingofcity: Removed

PR #497, opened 2021-04-21
[Issue-496] best practice of `optional`

### LINKED ISSUE #496: Best practice of `Optional`
**Problem description**

Java 8's Optional was mainly intended for return values from methods, and not for properties of Java classes, as described in [Optional in Java SE 8](https://blog.joda.org/2014/11/optional-in-java-se-8.html):

> > Of course, people will do what they want. But we did have a clear intention when adding this feature, and it was not to be a general purpose Maybe or Some type, as much as many people would have liked us to do so. **Our intention was to provide a limited mechanism for library method return types** where there needed to be a clear way to represent "no result", and using null for such was overwhelmingly likely to cause errors.
>
>The key here is the focus on use as a return type. The class is definitively not intended for use as a property of a Java Bean. Witness to this is that `Optional` [does not implement](http://mail.openjdk.java.net/pipermail/jdk8-dev/2013-September/003274.html) `Serializable`, which is generally necessary for widespread use as a property of an object.

---

This is referenced from [Is it a good practice to use Optional as an attribute in a class?](https://stackoverflow.com/questions/29033518/is-it-a-good-practice-to-use-optional-as-an-attribute-in-a-class)

**Problem location**

`src/main/java/io/pravega/connectors/flink/dynamic/table/FlinkPravegaDynamicTableSource.java`
`src/main/java/io/pravega/connectors/flink/dynamic/table/FlinkPravegaDynamicTableSink.java`

Use `@Nullable` instead.

thekingofcity:
**Change log description**

Java 8's Optional was mainly intended for return values from methods, and not for properties of Java classes, as described in [Optional in Java SE 8](https://blog.joda.org/2014/11/optional-in-java-se-8.html):

> > Of course, people will do what they want. But we did have a clear intention when adding this feature, and it was not to be a general purpose Maybe or Some type, as much as many people would have liked us to do so. **Our intention was to provide a limited mechanism for library method return types** where there needed to be a clear way to represent "no result", and using null for such was overwhelmingly likely to cause errors.
>
>The key here is the focus on use as a return type. The class is definitively not intended for use as a property of a Java Bean. Witness to this is that `Optional` [does not implement](http://mail.openjdk.java.net/pipermail/jdk8-dev/2013-September/003274.html) `Serializable`, which is generally necessary for widespread use as a property of an object.

---

This is referenced from [Is it a good practice to use Optional as an attribute in a class?](https://stackoverflow.com/questions/29033518/is-it-a-good-practice-to-use-optional-as-an-attribute-in-a-class)

**Purpose of the change**

Fixes #496 

**What the code does**

Use `@Nullable` instead of `Optional<T>` for class fields and function parameters.

**How to verify it**

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -554,13 +554,9 @@ public class FlinkPravegaReader<T>
             builder.append("scope=").append(scope).append(", ");
             builder.append("stream=").append(stream).append(", segments={");
 
-            readerGroup
-                    .getStreamCuts()
-                    .entrySet()
-                    .stream()
+            readerGroup.getStreamCuts().entrySet().stream()
                     .filter(e -> e.getKey().getStreamName().equals(stream) &&
-                            e.getKey().getScope().equals(scope))
-                    .findFirst()
+                            e.getKey().getScope().equals(scope)).findFirst()
                     .ifPresent(streamStreamCutEntry -> builder.append(streamStreamCutEntry.getValue().toString()));
 
             builder.append("}");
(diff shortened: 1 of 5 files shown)


crazyzhou (approved):
LGTM

PR #508, opened 2021-05-07
[issue-506] Improve the transactional writer

### LINKED ISSUE #506: Improve the transaction timeout handling in exactly-once writer
**Problem description**
There are several improvements that we can take for the transactional writer.
- Flink's `TwoPhaseCommitSinkFunction` class has offered some parameters like the timeout and warning ratio for implementers to take for recording abnormal transactions. The `FlinkPravegaWriter` can take usage of it.
- Add a `toString()` function for `PravegaTransactionState` so that logs can reflect the actual value of the state.
- In `checkTransactionStatus` call, there are some rare cases that hits an unknown transaction failure, we need to handle this exception internally.

**Problem location**
`FlinkPravegaWriter`

crazyzhou:
**Change log description**
- Add transaction timeout support
- Handle unknown transaction error and do not let it escape and fail the job
- Add `toString` method in `PravegaTransactionState`

**Purpose of the change**
Fixes #506 

**How to verify it**
Target to master, should cherry-pick to all `0.10` and `0.9` branch

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
@@ -148,6 +148,11 @@ public class FlinkPravegaWriter<T>
         this.enableWatermark = enableWatermark;
         this.enableMetrics = enableMetrics;
         this.writerIdPrefix = UUID.randomUUID().toString();
+
+        if (writerMode == PravegaWriterMode.EXACTLY_ONCE) {
+            super.setTransactionTimeout(txnLeaseRenewalPeriod);
+            super.enableTransactionTimeoutWarnings(0.8);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -288,6 +293,10 @@ protected void commit(PravegaTransactionState transaction) {
                     }
                 } catch (TxnFailedException e) {
                     log.error("{} - Transaction {} commit failed.", writerId(), txn.getTxnId());
+                } catch (RuntimeException e) {
+                    if (e.getMessage().contains("Unknown transaction")) {
+                        log.error("{} - Transaction {} not found.", writerId(), txn.getTxnId());
```
maddisondavid, line 298: Since the `eventStreamWriter` is a mock, is there a way we can test this by doing something like creating a mock that always returns Unknown Transaction?

crazyzhou: Yes, you are right. I was looking into integration test, and unit tests are easier to have such issue covered. Fixed with the recent commit

maddisondavid (approved):
LGTM

PR #518, opened 2021-05-24
[issue-517] Fix NPE issue for recovery job after enabling transactions

### LINKED ISSUE #517: Fix NPE issue when recovery job after enabling transactions
**Problem description**
When we turn EXACTLY_ONCE to On from Off for a running job, it will create a NPE and stop recovering. Connector should handle recovery when transaction information is missing.

**Problem location**
`FlinkPravegaWriter`

fyang86:

**Change log description**
When we turn EXACTLY_ONCE to On from Off for a running job, it will create a NPE and stop recovering.
Add conditions check when transaction information is missing.

**Purpose of the change**
Fix #517 

**What the code does**
Add checking conditions in `FlinkPravegaWriter`

**How to verify it**


src/main/java/io/pravega/connectors/flink/FlinkPravegaWriter.java
```diff
@@ -279,6 +279,9 @@ protected void preCommit(PravegaTransactionState transaction) throws Exception {
     protected void commit(PravegaTransactionState transaction) {
         switch (writerMode) {
             case EXACTLY_ONCE:
+                if (transaction.transactionId == null) {
```
crazyzhou, line 282: Please add some comment here like
```
// This may come from a job recovery from a non-transactional writer.
```

crazyzhou (changes requested):

crazyzhou (approved):
LGTM

PR #522, opened 2021-05-25
[Issue-520] Introduce dependency sorting in the checkstyle configurations

### LINKED ISSUE #520: Introduce dependency sorting in the checkstyle configurations
**Problem description**

Noticed that some imports are not sorted in the code base, this may be introduced from constant PRs, we need to fix this in all the 0.10 branches and try to add checks to avoid further similar issues.

According to https://checkstyle.sourceforge.io/config_imports.html#CustomImportOrder_Properties
```
<module name="CustomImportOrder">
  <property name="sortImportsInGroupAlphabetically" value="true"/>
</module>
```

**Problem location**
checkstyle configurations

thekingofcity:

**Change log description**

Noticed that some imports are not sorted in the code base, this may be introduced from constant PRs, we need to fix this in all the 0.10 branches and add checks to avoid further similar issues.

The desired ordering of the `import`.
* third party packages
* blank line
* java.*
* javax.*
* blank line
* static import

**Purpose of the change**

Fixes #520 

**What the code does**

There is no code change in this pr. Only the order of `import` is changed.

**How to verify it**

* all files have the right `import` ordering.
* `./gradlew checkstyleMain checkstyleTest` should pass.


crazyzhou (approved):
LGTM

PR #526, opened 2021-05-26
Issue 525 introduce dependency sorting in the checkstyle configurations in r0.10 flink1.11

### LINKED ISSUE #525: Introduce dependency sorting in the checkstyle configurations in r0.10-flink1.11
**Problem description**

Noticed that some imports are not sorted in the code base, this may be introduced from constant PRs, we need to fix this in all the 0.10 branches and try to add checks to avoid further similar issues.

According to https://checkstyle.sourceforge.io/config_imports.html#CustomImportOrder_Properties

```xml
<module name="CustomImportOrder">
  <property name="sortImportsInGroupAlphabetically" value="true"/>
</module>
```

**Problem location**

checkstyle configurations

thekingofcity:

**Change log description**

Noticed that some imports are not sorted in the code base, this may be introduced from constant PRs, we need to fix this in all the 0.10 branches and add checks to avoid further similar issues.

The desired ordering of the `import`.
* third party packages
* blank line
* javax.*
* java.*
* blank line
* static import

**Purpose of the change**

Fixes #525 

**What the code does**

* Change the order of `import`.
* fix the nonexisting RuntimeContextInitializationContextAdapters

**How to verify it**

* all files have the right `import` ordering.
* `./gradlew checkstyleMain checkstyleTest` should pass.
* `./gradlew clean build` should pass.


src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
```diff
@@ -345,8 +344,7 @@ public void cancel() {
 
     @Override
     public void open(Configuration parameters) throws Exception {
-        deserializationSchema.open(RuntimeContextInitializationContextAdapters.deserializationAdapter(
-                getRuntimeContext(), metricGroup -> metricGroup.addGroup("user")));
+        this.deserializationSchema.open(() -> getRuntimeContext().getMetricGroup().addGroup("user"));
```
crazyzhou, line 347: We also fixed a problem here when cherry-pick the commit from master branch

crazyzhou (approved):
LGTM

PR #527, opened 2021-05-26
[Issue-441] Fail to reset reader group with no checkpoint recovery

### LINKED ISSUE #441: Fail to reset reader group with no checkpoint recovery
**Problem description**
In `FlinkPravegaReaderRGStateITCase`, we have a test for no checkpoint recovery case. We expect the Pravega reader group should be reset to the beginning position with the help of the checkpoint hook when there is no checkpoint for recovery. 
However, https://issues.apache.org/jira/browse/FLINK-20222 changes the behavior we once had for this story. The change brings an enum of `OperatorCoordinatorRestoreBehavior` with a check for the `hook.reset()` . By default, it is a regional recovery strategy which sets it to `SKIP`, then the `hook.reset()` is not called. This patch is added in both 1.11.3 and 1.12.0, which causes the test case failure that blocks the connector integration.

**Problem location**
Flink's `CheckpointCoordinator`

thekingofcity:
**Change log description**

After the `CheckpointCoordinator` is fixed in Flink, we can have the test `FlinkPravegaReaderRGStateITCase` again.

**Purpose of the change**

Fixes #441 

**What the code does**

* Update Flink version to 1.12.4
* test `FlinkPravegaReaderRGStateITCase` again

**How to verify it**


crazyzhou (approved):
LGTM

PR #528, opened 2021-06-02
[Issue-490] Use maven publish instead of maven plugin in gradle

### LINKED ISSUE #490: Use maven-publish instead of maven plugin in gradle
**Problem description**

The current `maven` plugin is *deprecated* according to the [offical website](https://docs.gradle.org/6.8.3/userguide/maven_plugin.html) and is recommended to use the new `maven-publish` plugin.

**Problem location**

`./gradle/maven.gradle`

thekingofcity:
**Change log description**

The current `maven` plugin is *deprecated* according to the [offical website](https://docs.gradle.org/6.8.3/userguide/maven_plugin.html) and is recommended to use the new `maven-publish` plugin.

**Purpose of the change**

Fixes #490 

**What the code does**

* Use `maven-publish` plugin in gradle
* Upgrade gradle to `7.0.2`

**How to verify it**

The publish process and outcome should remain the same as the old plugin.


build.gradle
```diff
@@ -27,6 +26,7 @@ buildscript {
 }
 plugins {
     id "com.github.spotbugs"
+    id 'com.github.johnrengelman.shadow' version '7.0.0'
```
crazyzhou, line 29: Can we put it in `gradle/java.gradle` as before?

crazyzhou (approved):
LGTM

build.gradle
thekingofcity: fixed

Issue #507, opened 2021-05-07
Publish snapshot artifacts in GitHub Packages Registry but not in JCenter


crazyzhou:
**Problem description**
jcenter is being discontinued for public use and no more uploads would be possible.
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

Pravega needs to publish `-SNAPSHOT` artifacts somewhere so other projects can consume in-development versions during their development cycle

ref: https://github.com/pravega/pravega/issues/5984

**Problem location**
Build scripts and GH Actions

**Suggestions for an improvement**
One of potential alternatives for snapshot artifacts could be GitHub Packages Registry

PR #537, opened 2021-07-07
[issue-507] Publish snapshot artifacts in GitHub packages registry but not in Jcenter

thekingofcity:

**Change log description**
Changes the destination for snapshot artifacts from JCenter to GitHub Packages Registry.

**Purpose of the change**
Fixes pravega/flink-connectors#507 

**What the code does**
Snapshot action from build workflow is updated to publish to GPR:

- It uses auto-generated GITHUB_TOKEN token which is valid only in current repository
- By default publishes to owner/repository which is a repository where the action is executed
- Removes jcenter()/jfrog references from build scripts
- Exclude zookeeper 3.6.2 imported by pravega-standalone and wait for pravega/pravega#6129

**How to verify it**
Once PR is merged the next build on master branch must produce snapshot artifacts in GPR

**Known Limitations**

- Packages are not auto-deleted after some time. This will be addressed separately
- Using snapshot packages in other repository is not allowed without the authentication.

Limitation 2 can be handled in the following way:

- Create import token in pravega repository with read:packages scope. This can be shared with others
- In the other repository add the following to repositories list

```
maven {
           url "https://maven.pkg.github.com/pravega/pravega"
           credentials {
             username = "pravega"
             password = IMPORT_TOKEN // this is token created above 
           }
       }
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,7 +121,7 @@ jobs:
           key: ${{ github.run_id }}
 
       - name: Publish to Github Packages
-        run: ./gradlew publish -PpublishUrl=https://maven.pkg.github.com/${{github.repository}} -PpublishUsername=${{github.actor}} -PpublishPassword=${{secrets.GITHUB_TOKEN}}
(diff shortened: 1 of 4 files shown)


crazyzhou (approved):
LGTM

PR #549, opened 2021-07-27
[Issue-539] NullPointerPxception when there is no credentials in pravega client config

### LINKED ISSUE #539: NullPointerException when there's no credentials in Pravega client config
**Problem description**
When we call `getSchemaRegistryClientConfig` from class `SchemaRegistryUtils`, if there are no credentials in Pravega config's client config, we might pass a new `PravegaCredentialProvider` class with null `credentials` so that it could throw `NullPointerException` later when it calls `getMethod` inside `PravegaCredentialProvider`.

**Problem location**
`SchemaRegistryUtils.java`

thekingofcity:
**Change log description**

Prevent the `NullPointerException` if there is no credential after `getSchemaRegistryClientConfig` is called.

**Purpose of the change**

Fixes #539 

**What the code does**

Remove the authentication process if the pravegaConfig contains no credential.

**How to verify it**

Tests using SchemaRegistry should pass when the pravega is in the standalone mode and is configured without credentials.

> `FlinkPravegaSchemaRegistryReaderTestITCase.testReaderWithJsonRegistryDeserializer`.

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
+++ b/src/main/java/io/pravega/connectors/flink/FlinkPravegaReader.java
@@ -341,7 +341,7 @@ public class FlinkPravegaReader<T>
         while ((event = pravegaCollector.getRecords().poll()) != null) {
             synchronized (ctx.getCheckpointLock()) {
                 if (isEventTimeMode()) {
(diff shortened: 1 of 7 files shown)


crazyzhou (approved):
LGTM

PR #550, opened 2021-07-28
[issue-390] New Source API integration

### LINKED ISSUE #390: FLIP-27 new Source API integration
**Problem description**
Flink 1.11 has done with [FLIP-27](https://cwiki.apache.org/confluence/display/FLINK/FLIP-27%3A+Refactor+Source+Interface). This introduces a new source API with some explanation in the [official document](https://ci.apache.org/projects/flink/flink-docs-release-1.11/dev/stream/sources.html). 
This new API is marked `Beta` now, but we can start this work in order to catch up with the Flink community. This new API can have more ability to control of the reader side. 
In order to keep the compatibility, we will keep the old `FlinkPravegaReader`.

**Problem location**
Source

fyang86:
**Change log description**
This pull request brings the new Source API described in [FLIP-27](https://cwiki.apache.org/confluence/display/FLINK/FLIP-27%3A+Refactor+Source+Interface) since Flink 1.11.

**Purpose of the change**
- Fix #390 
- This change has brought a new `PravegaSource` as the Source implementation of Pravega without breaking the legacy Pravega source `FlinkPravegaReader`.

**What the code does**
- Introducing `PravegaSource` as the new Source implementation of Pravega.
- Introducing `PravegaSplitEnumerator` / `PravegaSplitReader` / `PravegaSourceReader` as main components.
- Introducing `PravegaSplit` as `SourceSplit` interface implementation.
- Introducing `PravegaFetcherManager` / `PravegaRecordEmitter` as main components of `PravegaSourceReader`.

**How to verify it**


src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
```diff
@@ -0,0 +1,200 @@
...
+        final String checkpointName = createCheckpointName(chkPtID);
+
+        LOG.info("Initiate checkpoint {}", checkpointName);
+        final CompletableFuture<Checkpoint> checkpointResult =
+                this.readerGroup.initiateCheckpoint(checkpointName, scheduledExecutorService);
+        try {
+            this.checkpoint = checkpointResult.get(5, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            LOG.error("Pravega checkpoint met error.", e);
+            this.checkpoint = null;
```
fpj, line 169: Is this the expected behavior in the case of failure, to set checkpoint to null rather than bubble the exception up?

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
```diff
@@ -0,0 +1,200 @@
...
+    // The scope name of the reader group.
+    private final String scope;
+
+    // The readergroup name to coordinate the parallel readers.
+    private final String readerGroupName;
+
+    // Pravega splits list.
+    private final List<PravegaSplit> splits;
+
+    // Pravega checkpoint for enymerator.
```
fpj, line 66: Typo: enumerator

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
```diff
@@ -0,0 +1,200 @@
...
+
+    // deal with the data recovery, it will call readerGroup::initiateCheckpoint to get a checkpoint for recovery.
+    // this call will be handled in another thread pool instead of the split enumerator thread.
+    @Override
+    public Checkpoint snapshotState(long chkPtID) throws Exception {
+        final String checkpointName = createCheckpointName(chkPtID);
+
+        LOG.info("Initiate checkpoint {}", checkpointName);
+        final CompletableFuture<Checkpoint> checkpointResult =
+                this.readerGroup.initiateCheckpoint(checkpointName, scheduledExecutorService);
```
fpj, line 164: Do I understand correctly that checkpointing the source is now done through the SplitEnumerator? Is the job manager now calling this method when the job is checkpointing?

src/main/java/io/pravega/connectors/flink/source/reader/PravegaSourceReader.java
```diff
@@ -0,0 +1,119 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+/**
+ * An Pravega implementation of {@link SourceReader}.
```
fpj, line 36:
```
A Pravega implementation of {@link SourceReader}.
```

src/main/java/io/pravega/connectors/flink/source/reader/PravegaRecordEmitter.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param deserializationSchema   The implementation to deserialize events from Pravega streams.
+     */
+    public PravegaRecordEmitter(DeserializationSchema<T> deserializationSchema) {
+        checkpointId = Optional.empty();
+        this.collector = new SimpleCollector<>();
+        this.deserializationSchema = deserializationSchema;
+    }
+
+    @Override
+    public void emitRecord(EventRead<ByteBuffer> record, SourceOutput<T> output, PravegaSplit state) throws Exception {
```
fpj, line 54: Where is this called from? `SourceReaderBase`?

fpj (changes requested):
It is looking good, thanks for proposing the changes.

src/main/java/io/pravega/connectors/flink/CheckpointSerializer.java
```diff
@@ -21,7 +21,7 @@
  * <p>The serializer currently uses {@link java.io.Serializable Java Serialization} to
  * serialize the checkpoint objects.
  */
-class CheckpointSerializer implements SimpleVersionedSerializer<Checkpoint> {
+public class CheckpointSerializer implements SimpleVersionedSerializer<Checkpoint> {
```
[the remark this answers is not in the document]
fyang86: Because the new `PravegaSource` needs to call it and it's in a different package

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
```diff
@@ -0,0 +1,200 @@
...
+        for (int i = 0; i < enumContext.currentParallelism(); i++) {
+            splits.add(new PravegaSplit(readerGroupName, i));
+        }
+        this.checkpoint = checkpoint;
+        this.scheduledExecutorService = Executors.newScheduledThreadPool(DEFAULT_CHECKPOINT_THREAD_POOL_SIZE);
+    }
+
+    // initiate reader group manager, reader group and reset the group to the checkpoint position if checkpoint isn't null
+    @Override
+    public void start() {
```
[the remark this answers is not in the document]
fyang86: yes it's just inherited, maybe the method is better to be called 'initialize' here

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
```diff
@@ -0,0 +1,200 @@
...
+
+    // This method is called when it tries to "regionally" recover from a failure to assign unassigned splits. We will
+    // throw an intentional exception here to trigger a global recovery since Pravega since Pravega
+    // doesn't have a mechanism to recover from partial failure. This will shutdown the reader group,
+    // recreate it and recover it from the latest checkpoint.
+    @Override
+    public void addSplitsBack(List<PravegaSplit> splits, int subtaskId) {
+        LOG.info("Call addSplitsBack {} : {}", splits.size(), subtaskId);
+
+        if (!isRecovered) {
```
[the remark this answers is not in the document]
fyang86: yes and no. `addSplitsBack` is called as the local recover logic and it won't deal with checkpoint. Currently one source reader is mapped to one split reader(Event Stream Reader). If it fails we have to reset the reader group from the checkpoint as Pravega's recovery mechanism. Hence we throw an exception here to trigger a global failure to reset the split enumerator since reader group is inside enumerator's scope.

src/test/java/io/pravega/connectors/flink/source/FlinkPravegaSourceTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.flink.test.util.AbstractTestBase;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.Timeout;
+
+import java.util.concurrent.TimeUnit;
+
+public class FlinkPravegaSourceTest extends AbstractTestBase {
```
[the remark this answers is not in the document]
fyang86: agreed, will add UTs later

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
```diff
@@ -0,0 +1,200 @@
...
+                    .maxOutstandingCheckpointRequest(this.readerGroupConfig.getMaxOutstandingCheckpointRequest())
+                    .groupRefreshTimeMillis(this.readerGroupConfig.getGroupRefreshTimeMillis())
+                    .disableAutomaticCheckpoints()
+                    .startFromCheckpoint(this.checkpoint)
+                    .build());
+        }
+    }
+
+    @Override
+    public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname) {
```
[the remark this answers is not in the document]
fyang86: agreed, since we never called `sendSplitRequest` from source reader, this method will never be called here

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
```diff
@@ -0,0 +1,200 @@
...
+        // the Pravega source pushes splits eagerly, rather than act upon split requests
+    }
+
+    // check with the current parallelism and update the assignment (always one-to-one mapping)
+    @Override
+    public void addReader(int subtaskId) {
```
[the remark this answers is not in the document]
fyang86: one `PravegaSourceReader` is mapped to one unique subtaskId, also one `PravegaSourceReader` contains one `PravegaSplitReader` which reads one `PravegaSplit`

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
fyang86: yes. It is called when job us checkpointing.

src/main/java/io/pravega/connectors/flink/source/reader/PravegaRecordEmitter.java
fyang86: It is called in method `pollNext` in `SourceReaderBase`

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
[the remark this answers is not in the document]
fpj: Should we throw unsupported operation then?

https://docs.oracle.com/javase/8/docs/api/java/lang/UnsupportedOperationException.html

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    @Override
+    public void close() throws IOException {
+        readerGroup.close();
+        readerGroupManager.close();
+    }
+
+    // This method is called when there is an attempt to locally recover from a failure to assign unassigned splits.
+    // It throws an intentional exception here to trigger a global recovery since Pravega since Pravega
```
fpj, line 188: Typo: `since Pravega` twice.

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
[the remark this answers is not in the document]
fpj: I'm not sure I understand why it needs to reset to a checkpoint. Could we make the event reader offline instead? Perhaps I'm not understanding what is expected to fail here.

src/main/java/io/pravega/connectors/flink/source/reader/PravegaSourceReader.java
```diff
@@ -0,0 +1,125 @@
...
+        if (checkpointId.isPresent()) {
+            inputStatus = InputStatus.NOTHING_AVAILABLE;
+            LOG.trace("Source reader status: {}", inputStatus);
+        }
+        return inputStatus;
+    }
+
+    @Override
+    public List<PravegaSplit> snapshotState(long checkpointId) {
+        // Pravega doesn't support partial failover recover so just return null here for Source Reader
```
fpj, line 102: Same question about partial failure, could we make the faulty reader offline for partial recovery?

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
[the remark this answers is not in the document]
fpj: I'll follow up below.

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
```diff
@@ -0,0 +1,206 @@
...
+    public void addReader(int subtaskId) {
+        if (enumContext.registeredReaders().size() == enumContext.currentParallelism() && splits.size() > 0) {
+            LOG.debug("Adding reader {} to PravegaSplitEnumerator for reader group: {}/{}.",
+                    subtaskId,
+                    scope,
+                    readerGroupName);
+            int numReaders = enumContext.registeredReaders().size();
+            Map<Integer, List<PravegaSplit>> assignment = new HashMap<>();
+            for (int i = 0; i < numReaders; i++) {
+                assignment.put(i, Collections.singletonList(splits.get(i)));
```
fpj, line 155: Could you please document as a comment how this assignment is being performed? The `subtaskId` parameter isn't used for the assignment, and we seem to be assigning `i` to `i` here, which seems to match the desired mapping of subtask ID to one split. I'd like to see some explanation as of why the assignment is being performed this way.

fpj (changes requested):

src/main/java/io/pravega/connectors/flink/source/reader/PravegaSourceReader.java
crazyzhou: This also was my initial idea and I also tried but failed. It takes me some time and also discussions to Pravega client team to understand the gap here. I'll try to explain here.

For partial failure recovery, it would be handled by Flink that recover part of the DAG that connected to the faulty reader on the lineage and recover this region to the last checkpoint(more detail on [FLIP-1](https://cwiki.apache.org/confluence/display/FLINK/FLIP-1+%3A+Fine+Grained+Recovery+from+Task+Failures)). This means that a single reader needs to have the ability to rewind back to the last checkpoint position.

For Pravega API, it is not working like that. When the client calls the `readerOffline`, the reader group state manager would reassign the segments owned by the reader, and the other online readers would take over and continue reading. It behaves too smart. If we do the `readerOffline` with the FLIP-1 recovery, we would find that the downstream operators in Flink would have the state stored in Flink checkpoint, while the source has already go beyond the Pravega checkpoint, which leads to duplicates.

src/main/java/io/pravega/connectors/flink/CheckpointSerializer.java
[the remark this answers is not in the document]
fyang86: fixed this in #630

src/test/java/io/pravega/connectors/flink/source/FlinkPravegaSourceTest.java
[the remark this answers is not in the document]
fyang86: already added unit test for sourceReader, splitReader and enumerator

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
fyang86: I have changed the implementation here to perform the assignment once this method is called by assigning the Pravega Split with incoming `subtaskId` to Source Reader to make it more concise and more readable. Please help rereview it.

Also let me explain the original design briefly. It was intended to assign all splits at a time when all readers are registered. Here it won't go into the `if` statement until the number of register readers equals `currentParallelism`. Then we can perform the assignment in the `for` loop because the numbering of `subtaskId` starts from 0 and goes up to parallelism-1 according to [this](https://nightlies.apache.org/flink/flink-docs-stable/api/java/org/apache/flink/api/common/functions/RuntimeContext.html#getIndexOfThisSubtask--), which can map to the index `i`.

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
fyang86: fixed

src/main/java/io/pravega/connectors/flink/source/PravegaSource.java
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * The Source implementation of Pravega. Please use a {@link PravegaSourceBuilder} to construct a {@link
+ *  PravegaSource}. The following example shows how to create a PravegaSource emitting records of <code>
+ *  Integer</code> type.
+ *
+ * <pre>{@code
+ * PravegaSource<Integer> pravegaSource = PravegaSource.<Integer>builder()
+ *                     .forStream(streamName)
+ *                     .enableMetrics(false)
+ *                     .withPravegaConfig(PRAVEGA_CONFIG)
```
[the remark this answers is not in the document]
fyang86: Typo here because it was straight copied from a unit test. It should not be a constant, will fix it

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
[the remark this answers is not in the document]
fyang86: agreed with this, we should throw an `UnsupportedOperationException` to indicate the method would be considered improper or invalid

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
fyang86: Yes we should bubble the exception up here rather than returning null and leave it for flink to handle it. 
Also `checkpoint` should not be set to null because it represents the 'latest' checkpoint of the enumerator, thus, should keep unchanged when initiating new checkpoint fails.

src/main/java/io/pravega/connectors/flink/source/enumerator/PravegaSplitEnumerator.java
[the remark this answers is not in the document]
fyang86: after discussion with Brian, we think that we don't need to throw an exception here. 

The `handleSplitRequest` method is called in `handleEventFromOperator` when an coming event is a `RequestSplitEvent` instance in `SourceCoordinator`, any exception thrown will results in a full job failure and recovery which is not an expected behavior. So we'll do dothing here if the event coming to Enumerator is a `RequestSplitEvent` instance by mistake.

src/main/java/io/pravega/connectors/flink/source/reader/PravegaSourceReader.java
fpj: I see, you are saying that partial recovery expects recovery of the faulty reader _r_ to roll back to a specific point, which is the last checkpoint position for the segments previously owned by _r_. if that's the case, then could we pass a  `Position` object as part of the `readerOffline` call that corresponds to that position? I guess the follow up question is how we build such a `Position` object. I wonder if it is possible to derive a position from the checkpoint object.

crazyzhou: Not just that, after the reader offline, the other readers will take over the segment and continue reading, this would cause duplicates even if we can roll back the failed reader to the position. Therefore, it is not a valid partial recovery.

fpj: I'll approve to unblock the work, but we should have an issue to discuss the possibility of enabling partial recovery. I'm concerned that it will be perceived as a disadvantage of the model that Pravega is following.

fpj (approved):

crazyzhou (approved):
LGTM, but still having issue #663

src/main/java/io/pravega/connectors/flink/source/reader/PravegaSourceReader.java
crazyzhou: I have created #663 to track this issue, but it is true that it is now a disadvantage of the model

PR #553, opened 2021-08-06
[issue-552] Support SpecificRecordBase in DeserializerFromSchemaRegistry

### LINKED ISSUE #552: Support SpecificRecordBase in DeserializerFromSchemaRegistry
**Problem description**
In `DeserializerFromSchemaRegistry`, currently we create an `AvroGenericDeserializer` when type class is `GenericRecord` or its subclasses, which will ignore the circumstance where type class is `SpecificRecordBase` since it is also a subclass of `GenericRecord`.

**Problem location**
`DeserializerFromSchemaRegistry`

fyang86:

**Change log description**
Add support for avro format `SpecificRecordBase` in `DeserializerFromSchemaRegistry`.

**Purpose of the change**
Fix #552 

**What the code does**
Change if order in `DeserializerFromSchemaRegistry`.
Update unit test `FlinkPravegaSchemaRegistryReaderTestITCase`

**How to verify it**


src/test/java/io/pravega/connectors/flink/FlinkPravegaSchemaRegistryReaderTestITCase.java
```diff
@@ -51,19 +53,62 @@ public MyTest(String a) {
         }
     }
 
+    @SuppressWarnings("checkstyle:ConstantName")
```
crazyzhou, line 56: We don't need to keep the constant name the same with the avro-generated class, a constant named `SCHEMA` would be enough. the class name can also be simplified to `User` as well.
Another path is that we keep this class real avro generated, we need to have a `.avsc` file in the `resources` folder and let avro generate the class. See 

https://github.com/apache/flink/blob/master/flink-formats/flink-avro-glue-schema-registry/src/test/java/resources/avro/user.avsc
https://github.com/apache/flink/blob/master/flink-formats/flink-avro-glue-schema-registry/src/test/java/org/apache/flink/formats/avro/glue/schema/registry/User.java

Both approaches are good to me.

src/test/java/io/pravega/connectors/flink/FlinkPravegaSchemaRegistryReaderTestITCase.java
```diff
@@ -51,19 +53,62 @@ public MyTest(String a) {
...
             new SchemaRegistryUtils(SETUP_UTILS, SchemaRegistryUtils.DEFAULT_PORT);
 
-    private static final Schema SCHEMA = SchemaBuilder
-            .record("MyTest")
-            .fields()
-            .name("a")
-            .type(Schema.create(Schema.Type.STRING))
-            .noDefault()
-            .endRecord();
+    private static final Schema SCHEMA = new Schema.Parser().parse(
```
crazyzhou, line 108: Can we use `SpecificRecordTestClass::SCHEMA` here? Looks like they are the same

crazyzhou (changes requested):

src/test/java/io/pravega/connectors/flink/FlinkPravegaSchemaRegistryReaderTestITCase.java
fyang86: agreed, will fix this

crazyzhou (approved):
LGTM

PR #555, opened 2021-08-09
[issue-554] Update HEADER

### LINKED ISSUE #554: Update the license header
**Problem description**
Update license header corresponding to https://github.com/pravega/pravega/pull/5807

**Problem location**
The whole repository

fyang86:

**Change log description**
* Updates HEADER file to reflect copyright changes corresponding to https://github.com/pravega/pravega/pull/5807
* Add checkstyle to validate the correct copyright.

**Purpose of the change**
Fix #554 

**What the code does**
No code change

**How to verify it**


NOTICE
```diff
@@ -1 +1 @@
-Copyright (c) 2017-2021 Dell Inc., or its subsidiaries. All Rights Reserved.
+Copyright Pravega Authors.
```
crazyzhou, line 1: @fpj @derekm  Should we leave the dell copyright as pravega repo here? https://github.com/pravega/pravega/blob/master/NOTICE#L2

fpj: We have replaced the all in the `pravega/pravega` repo, and the idea is indeed that we replace them across all repositories. @derekm did say that he would create issues across repositories so that we could change it across the org. @derekm?

derekm: Yeah, we should preserve old copyrights in the NOTICE file and also include years. For example:
https://github.com/pravega/pravega/blob/master/NOTICE

In file headers, we have permission from Dell Legal to replace old copyrights. For example:
https://github.com/pravega/pravega/blob/master/segmentstore/server/src/main/java/io/pravega/segmentstore/server/SegmentContainerManager.java#L2

fyang86:
reopen PR in https://github.com/pravega/flink-connectors/pull/583 becuase this branch was kind of screwed

PR #562, opened 2021-08-17
[issue-561] Fix exception types for Pravega catalog

### LINKED ISSUE #561: Fix type of some exceptions in Pravega catalog
**Problem description**
In `PravegaCatalog`, we now throw an `UnsupportedOperationException` for every method that we don't support. However, this may break some parts of catalog. For example, when the flink `FunctionCatalog` calls `catalog.getFunction()`, it will ignore the exception if it is a `FunctionNotExistException`. Therefore we need to change some exception types in order to not breaking the catalog.

**Problem location**
`PravegaCatalog`

fyang86:

**Change log description**
Change some exceptions and return statements in `PravegaCatalog` to not break the catalog function.

**Purpose of the change**
Fix #561 

**What the code does**
- For all list methods, return an empty list instead of throwing an `UnsupportedOperationException`.
- For all exists methods, return `false` instead of throwing an `UnsupportedOperationException`.
- For `getFunction`, throw a `FunctionNotExistException` instead of `UnsupportedOperationException`.
- For `getPartition`, throw a `PartitionNotExistException` instead of `UnsupportedOperationException`.

**How to verify it**

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/table/catalog/pravega/PravegaCatalog.java
+++ b/src/main/java/io/pravega/connectors/flink/table/catalog/pravega/PravegaCatalog.java
@@ -259,7 +259,7 @@ public class PravegaCatalog extends AbstractCatalog {
 
     @Override
     public List<String> listViews(String databaseName) throws DatabaseNotExistException, CatalogException {
-        throw new UnsupportedOperationException();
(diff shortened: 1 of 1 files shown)


crazyzhou (approved):
LGTM

PR #573, opened 2021-09-01
[issue-572] Use get method of CompletableFuture to prevent blocking calls

### LINKED ISSUE #572: Use `get` method of `CompletableFuture` instead of scheduling a new cancel task
**Problem description**
When initiating checkpoint for Pravega source, we currently are scheduling a cancel task if it triggers timeout. Will just use `CompletableFuture::get()` instead.

**Problem location**
`ReaderCheckpointHook`

fyang86:

**Change log description**
Use `CompletableFuture::get` to prevent blocking calls when it triggers timeout instead of scheduling a new cancel task.

**Purpose of the change**
Fix #572 

**What the code does**
Call `CompletableFuture::get` method, if it throws a timeout exception cancel the future.

**How to verify it**


crazyzhou:
As this method should be non-blocking for Flink, using `get` is not a good idea here, closing this for now.

PR #633, opened 2021-12-22
fix npe when close pravega batch client factory in batch mode

chenlei677:
Purpose of the change

What the code does

How to verify it


crazyzhou (changes requested):
Thanks for the contribution, two small suggestions here:
1. Create an issue first, and committers would label them before coding the bugfixes
2. As this is a general issue, the PR should go against `master` branch and then cherry-pick back to Flink 1.12 branch to fix your issue.

Issue #635, opened 2021-12-27
fix npe when close pravega batch client factory in batch mode


chenlei677:
**Problem description**
Fix npe when close pravega batch client factory in batch mode
java.lang.NullPointerException
 at io.pravega.connectors.flink.FlinkPravegaInputFormat.close(FlinkPravegaInputFormat.java:178)
 at org.apache.flink.streaming.api.functions.source.InputFormatSourceFunction.run(InputFormatSourceFunction.java:105)
 at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:110)
 at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:66)
 at org.apache.flink.streaming.runtime.tasks.SourceStreamTask$LegacySourceFunctionThread.run(SourceStreamTask.java:263)
 Suppressed: java.lang.NullPointerException
  at io.pravega.connectors.flink.FlinkPravegaInputFormat.close(FlinkPravegaInputFormat.java:178)
  at org.apache.flink.streaming.api.functions.source.InputFormatSourceFunction.close(InputFormatSourceFunction.java:120)
  at org.apache.flink.api.common.functions.util.FunctionUtils.closeFunction(FunctionUtils.java:41)
  at org.apache.flink.streaming.api.operators.AbstractUdfStreamOperator.dispose(AbstractUdfStreamOperator.java:117)
  at org.apache.flink.streaming.runtime.tasks.StreamTask.disposeAllOperators(StreamTask.java:797)
  at org.apache.flink.streaming.runtime.tasks.StreamTask.runAndSuppressThrowable(StreamTask.java:776)
  at org.apache.flink.streaming.runtime.tasks.StreamTask.cleanUpInvoke(StreamTask.java:691)
  at org.apache.flink.streaming.runtime.tasks.SourceStreamTask.cleanUpInvoke(SourceStreamTask.java:180)
  at org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:595)
  at org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:758)
  at org.apache.flink.runtime.taskmanager.Task.run(Task.java:573)
  at java.lang.Thread.run(Thread.java:748)

**Problem location**
io.pravega.connectors.flink.FlinkPravegaInputFormat.close

**Suggestions for an improvement**

chenlei677:
@crazyzhou Hello, I have created a new Issue, please help me to take a look. Thanks

PR #637, opened 2022-01-04
fix npe when close pravega batch client factory in batch mode

chenlei677:
Signed-off-by: chenlei677 <chenlei677@jd.com>

**Change log description**
#635 

**Purpose of the change**
When the batch task is executed, avoid the null pointer exception

**What the code does**

**How to verify it**


crazyzhou (commented):
@chenlei677 Thank you for the fix. Could you edit the PR to merge into the master branch? ( all changes to `r0.10` branch are cherry-picked from the master branch).

PR #638, opened 2022-01-04
[issue-635] Fix null pointer when closing pravega batch client factory

chenlei677:
Change log description
#635 

Purpose of the change
When the batch task is executed, avoid the null pointer exception

What the code does

How to verify it


crazyzhou (approved):
LGTM

PR #641, opened 2022-01-14
Update README.md

HeChuanXUPT:
fix version, links error


crazyzhou:
Thanks for the contribution, but we have addressed it in #643

PR #646, opened 2022-01-18
[issue-645] Update the year 2022 in copyright

### LINKED ISSUE #645: Update the year 2022 in copyright
**Problem description**
`NOTICE` file needs to be update to reflect the current year 2022

**Problem location**
`NOTICE`

Change the range to 2021-2022

fyang86:

**Change log description**
- Update `NOTICE` file to reflect the current year 2022

**Purpose of the change**
Fix #645 

**What the code does**
- Update year range in `NOTICE` file

**How to verify it**
No special verification is required, build should pass though

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,2 +1,2 @@
 Copyright (c) 2021-2022 Pravega Authors.
-Copyright (c) 2017-2022 Dell Inc., or its subsidiaries. All Rights Reserved.
+Copyright (c) 2017-2021 Dell Inc., or its subsidiaries. All Rights Reserved.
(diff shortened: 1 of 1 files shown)


NOTICE
```diff
@@ -1,2 +1,2 @@
-Copyright (c) 2021 Pravega Authors.
-Copyright (c) 2017-2021 Dell Inc., or its subsidiaries. All Rights Reserved.
\ No newline at end of file
+Copyright (c) 2021-2022 Pravega Authors.
+Copyright (c) 2017-2022 Dell Inc., or its subsidiaries. All Rights Reserved.
```
crazyzhou, line 2: It should still keep `2021` this line because dell copyrights are just used in the past. @derekm Please confirm if I am wrong here.

crazyzhou (changes requested):

NOTICE
derekm: Yes, that's correct @crazyzhou, the Dell Inc. copyright doesn't need to be maintained any longer. "Pravega Authors" is the plural copyright holder under CNCF.

crazyzhou (approved):
LGTM

Issue #651, opened 2022-02-10
Upgrade Flink to 1.15


thekingofcity:
**Problem description**
Flink will release 1.15 in the coming months and we need to support it.

**Problem location**
*All over the place.*

**Suggestions for an improvement**
* [FLINK-23986](https://issues.apache.org/jira/browse/FLINK-23986) Add support for opting-out of Scala
* [FLINK-24427](https://issues.apache.org/jira/browse/FLINK-24427) Hide Scala in flink-table-planner from API
* [FLINK-25276](https://issues.apache.org/jira/browse/FLINK-25276) FLIP-203: Support native and incremental savepoints
* [FLINK-25555](https://issues.apache.org/jira/browse/FLINK-25555) FLIP-191: Extend unified Sink interface to support small file compaction

PR #655, opened 2022-02-15
[Issue-653] FLIP-191 extend unified sink interface to support small file compaction

### LINKED ISSUE #653: FLIP-191 Extend unified Sink interface to support small file compaction
**Problem description**
This FLIP upgrade the new unified Sink from V1 to V2 introduced in #460.

**Problem location**
Previous sink.

thekingofcity:
**Change log description**
This FLIP upgrade the new unified Sink from V1 to V2.

**Purpose of the change**
Fix #653 

**What the code does**
Upgrade sink code to be compatible with V2.

**How to verify it**
`.\gradlew clean build` passes.


crazyzhou (changes requested):
After switching to sink v2, we still want to expose a single `PravegaSink` interface to the end user.
The object that `PravegaSinkBuilder` builds is not a `PravegaSink`, this not only violates the builder design pattern, but also brings additional confusion of the build output to the end users.

crazyzhou:
Since #651 is fixed, this PR can target the latest master branch.

src/main/java/io/pravega/connectors/flink/sink/PravegaSinkBuilder.java
```diff
@@ -47,6 +46,10 @@
     PravegaSinkBuilder() {
     }
 
+    static <T> PravegaSinkBuilder<T> builder() {
```
crazyzhou, line 49: This builder function should be in the `PravegaSink`, so that we can keep the API compatibility.

thekingofcity: Follow the builder pattern in source.

src/main/java/io/pravega/connectors/flink/sink/StreamNameGauge.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package io.pravega.connectors.flink.sink;
+
+import org.apache.flink.metrics.Gauge;
+
+/**
+ * Gauge for getting the fully qualified stream name information.
+ */
+public class StreamNameGauge implements Gauge<String> {
```
crazyzhou, line 23: As we do not have metrics support in this PR, this class can be removed.

crazyzhou (changes requested):

src/main/java/io/pravega/connectors/flink/sink/StreamNameGauge.java
thekingofcity: removed

crazyzhou (approved):
LGTM

PR #670, opened 2022-04-01
[Issue-668] Manage all the source and sink configurations in Flink ConfigOptions

### LINKED ISSUE #668: Manage all the source and sink configurations in Flink ConfigOptions
**Problem description**
It could be better if we have a central place to store all the client configurations, but now we are separating them in each of the source/sink builder.
This is not an ideal way because these configurations are not clearly exposed to the end users.

Now in order not to break the compatibility, we could keep the legacy source and sink unchanged, but enhance our new source and sink inplementation to use the new config system.

**Problem location**
New source and sink

thekingofcity:
Signed-off-by: thekingofcity <3353040+thekingofcity@users.noreply.github.com>

**Change log description**
It could be better if we have a central place to store all the client configurations, but now we are separating them in each of the source/sink builder.
This is not an ideal way because these configurations are not clearly exposed to the end users.

**Purpose of the change**
Fix #668 

**What the code does**
Use a new `PravegaOptions` to set and get Pravega client config.

**How to verify it**
`.\gradlew clean build` should pass.

### DIFF
--- a/src/test/java/io/pravega/connectors/flink/source/FlinkPravegaSourceITCase.java
+++ b/src/test/java/io/pravega/connectors/flink/source/FlinkPravegaSourceITCase.java
@@ -131,7 +131,12 @@ public class FlinkPravegaSourceITCase extends AbstractTestBase {
             // the Pravega reader
             final PravegaSource<Integer> pravegaSource = PravegaSource.<Integer>builder()
                     .forStream(streamName)
-                    .withPravegaClientConfig(SETUP_UTILS.getPravegaClientConfig())
+                    .withControllerURI(SETUP_UTILS.getControllerUri().toString())
+                    .withDefaultScope(SETUP_UTILS.getScope())
+                    .withUsername(SetupUtils.getUsername())
+                    .withPassword(SetupUtils.getPassword())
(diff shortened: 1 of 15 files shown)


src/test/java/io/pravega/connectors/flink/source/FlinkPravegaSourceITCase.java
```diff
@@ -131,8 +131,7 @@ private static void runTest(
             // the Pravega reader
             final PravegaSource<Integer> pravegaSource = PravegaSource.<Integer>builder()
                     .forStream(streamName)
-                    .enableMetrics(false)
-                    .withPravegaConfig(SETUP_UTILS.getPravegaConfig())
+                    .withPravegaClientConfig(SETUP_UTILS.getPravegaClientConfig())
```
crazyzhou, line 134: In FLIP-27 tests, we should set the config one by one instead of using a `Configuration` here, and even I think it's fine for now to remove the `withPravegaClientConfig()`. We can add it back when we need this to integrate the table API in the future.

crazyzhou (changes requested):

Issue #460, opened 2021-03-03
FLIP-143 new Sink API integration


crazyzhou:
**Problem description**
Since Flink 1.12, Flink has introduced a new interface for sink functions in [FLIP-143](https://cwiki.apache.org/confluence/display/FLINK/FLIP-143%3A+Unified+Sink+API)
This new interface needs to be integrated, this should be relatively easy because a standard `TwoPhaseCommitSinkFunction` is used by us after #377 is fixed.

**Problem location**
Writer

**Suggestions for an improvement**

PR #712, opened 2022-12-20
[Issue-705] Implement sink metrics

### LINKED ISSUE #705: Implement metrics for the Sink API
**Problem description**
With #460 , the new sink interface needs to expose certain metrics.

**Problem location**
Sink API

Add numRecordsOut, numBytesOut, refer to current Flink code and https://github.com/apache/flink/commit/545fd7a9e306b2fb32532701bdc80314d315c7c6

fyang86:
**Change log description**
With #460 , the new sink interface needs to expose certain metrics.
We only add `numRecordsOut` and `numRecordsOutError` metrics. `numBytesOut` is not available because Pravega itself does not have metrics on writer side.

**Purpose of the change**
Fix #705 

**What the code does**
Implement `numRecordsOut` for both event writer and transaction writer, implement `numRecordsOutError` for event writer.

**How to verify it**
Such metrics should be able to view through Flink UI or Flink RSET API

### DIFF
--- a/src/main/java/io/pravega/connectors/flink/sink/PravegaEventWriter.java
+++ b/src/main/java/io/pravega/connectors/flink/sink/PravegaEventWriter.java
@@ -28,6 +28,7 @@ import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.serialization.SerializationSchema;
 import org.apache.flink.api.connector.sink2.Sink;
 import org.apache.flink.api.connector.sink2.SinkWriter;
+import org.apache.flink.metrics.Counter;
 import org.apache.flink.util.ExceptionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -92,6 +93,15 @@ public class PravegaEventWriter<T> implements SinkWriter<T> {
     // The writer id
     private final String writerId;
 
+    // flag to enable/disable metrics
+    private final boolean enableMetrics;
+
+    // The total number of output records
+    private final Counter numRecordsOutCounter;
+
+    // The total number of records failed to send
+    private final Counter numRecordsOutErrorsCounter;
+
     /**
      * A Pravega non-transactional writer that handles {@link PravegaWriterMode#BEST_EFFORT} and
      * {@link PravegaWriterMode#ATLEAST_ONCE} writer mode.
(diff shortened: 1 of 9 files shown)


src/main/java/io/pravega/connectors/flink/sink/PravegaEventWriter.java
```diff
@@ -92,6 +93,15 @@
     // The writer id
     private final String writerId;
 
+    // flag to enable/disable metrics
+    private final boolean enableMetrics;
```
crazyzhou, line 97: Normally `enableMetrics` flag is to mark if we need to connect to the internal client metric system, which may have some impact on the client behavior and performance. As the current counter is done on connector side, we can by default enable metrics and remove the flag.

src/test/java/io/pravega/connectors/flink/sink/PravegaSinkWriterITCase.java
```diff
@@ -0,0 +1,213 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.nio.ByteBuffer;
+import java.util.OptionalLong;
+
+import static io.pravega.connectors.flink.sink.PravegaSinkBuilder.DEFAULT_TXN_LEASE_RENEWAL_PERIOD_MILLIS;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class PravegaSinkWriterITCase {
```
crazyzhou, line 50: We need to think how to test the error counter as well, now it's not tested.

crazyzhou (approved):
LGTM

src/main/java/io/pravega/connectors/flink/sink/PravegaTransactionalWriter.java
```diff
@@ -159,6 +164,7 @@ public void write(T element, Context context) throws IOException, InterruptedExc
         } catch (TxnFailedException | AssertionError e) {
             throw new IOException(e);
         }
+        numRecordsOutCounter.inc();
```
RaulGracia, line 167: If a transaction is aborted, this would have accounted the records of the failed transactions as successful records written. Is that the intended goal here?

fyang86: In the sink API a transaction is only aborted when we close the writer. When the transaction fails, it should throw an exception here so the counter won't increase.

PR #718, opened 2023-07-13
[issue-717] Let Pravega internal thread pool deal with checkpoint

### LINKED ISSUE #717: Let Pravega internal thread pool to deal with checkpoint
**Problem description**
Pravega has introduced a different way to initiate a checkpoint by using its internal thread pool in https://github.com/pravega/pravega/issues/4816

As the thread pool in Pravega has more control, it's better to switch to use that.

**Problem location**
`ReaderCheckpointHook`

crazyzhou:
**Change log description**
Use `ReaderGroup::initiateCheckpoint(checkpointName)` to perform the checkpoint

**Purpose of the change**
Fixes #717 

**What the code does**
Update the method
Update the Pravega version

**How to verify it**

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -40,7 +40,7 @@ testcontainersVersion=1.15.3
 
 # Version and base tags can be overridden at build time.
 connectorVersion=0.13.0-SNAPSHOT
-pravegaVersion=0.13.0-3151.28c485f-SNAPSHOT
+pravegaVersion=0.13.0-3181.f1783dd-SNAPSHOT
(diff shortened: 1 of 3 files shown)


ShwethaSNayak (approved):
LGTM, Looks like build is failed.

tkaitchuck (approved):

crazyzhou:
The test failure is because of this code change, looks like the change introduce a deadlock in the client internal thread in the read group. We cannot make such change.

PR #720, opened 2023-09-07
[issue-719] Explicitly cancel outstanding checkpoints when checkpoints get stuck

### LINKED ISSUE #719: Explicitly cancel outstanding checkpoints when checkpoints get stuck
**Problem description**
In some of the negative test cases, there will be some situation that the old checkpoints cannot complete its future, causing itslef unable to be released from the internal queue of Pravega outstanding checkpoints.

In https://github.com/pravega/pravega/issues/7147, Pravega introduced a new `cancelOutstandingCheckpoints` call.
The connector can detect this exception and explicitly call this method

**Problem location**
Checkpoint hook

Catch the `MaxNumberOfCheckpointsExceededException` and handle the call

crazyzhou:
**Change log description**
Add a `cancelOutstandingCheckpoints` call when the checkpoint hits `MaxNumberOfCheckpointsExceededException`

**Purpose of the change**
Fixes #719

**What the code does**
Add a `cancelOutstandingCheckpoints` call when the checkpoint hits `MaxNumberOfCheckpointsExceededException`

**How to verify it**

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -40,7 +40,7 @@ testcontainersVersion=1.15.3
 
 # Version and base tags can be overridden at build time.
 connectorVersion=0.13.0-SNAPSHOT
-pravegaVersion=0.13.0-3151.28c485f-SNAPSHOT
+pravegaVersion=0.13.0-3184.23d8ef6-SNAPSHOT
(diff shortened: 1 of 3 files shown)


anishakj (approved):
LGTM

PR #723, opened 2023-09-28
Add Protobuf Support

Welkin-Y:
**Change log description**

**Purpose of the change**
Add Protobuf Support

**What the code does**
Enable convertion bewtween Flink RowData and Pravega byte array in Protobuf format

**How to verify it**


crazyzhou (approved):
Code change looks good, will follow up on the tests, docs and use case next

